PR #6, opened 2020-12-23
fix_include_problem_lv_refr_h

liebman:
using v7.8 and later needs to remove the leading "lvgl" from the include path for `lv_refr.h`.  This PR uses LV_LVGL_H_INCLUDE_SIMPLE as the condition the same way `lvgl.h` is referenced.


C47D:
Thanks for the fix, I have it locally but forgot to push it.

PR #9, opened 2020-12-24
XPT2046 optionally (Kconfig option) validate a touch with pressure, not just IRQ low

### LINKED ISSUE #7: XPT2046 improvement discussion
I've been having an issue with this input driver with spurious input events. This is where the IRQ line goes low for ~ 8us, measured with a scope.  Looking at other implementations I see that some compute the Z (or press pressure). Maybe this should only return a new input based on some Z threshold.

Now the datasheet is very confusing on this subject and gives formula that require knowing the X and or Y resistance values. (currently I'm playing with what [this library does](https://github.com/PaulStoffregen/XPT2046_Touchscreen/blob/aef8ec703e2a04617a331e127d441b58e549c01a/XPT2046_Touchscreen.cpp#L131-L135).)

Any thoughts?

liebman:
Sometimes IRQ is very noisy an prevents things like idle time detection. This PR will check to insure there is pressure detected on the panel, not just the noisy IRQ signal. Disabled by default in Kconfig.
Fixes #7


liebman:
I've added an additional config option to use pressure only and ignore IRQ (not even configure the pin).

C47D:
Hi @liebman, all looks good. Do you think is it possible to move the touch detection code into a new function that returns 0 when there's no touch detected and another positive number when there's touch detected? Something like `xpt2048_is_press_detected`?

liebman:
@C47D is there a specific use case for this?  (your asking for this to be non-static I assume)

C47D:
> EDIT: or maybe this is just to make the logic more readable.... Can it return a bool?

Yes, to make the logic more readable, and yes, it can return a bool or an enum, I find enums be more readables.

```c
typedef enum {
TOUCH_NOT_DETECTED = 0,
TOUCH_DETECTED = 1,
} xpt2046_touch_detected;
```

liebman:
@C47D I'm looking for a better name for the `Z_MIN` constant.  Any ideas?  (as usual it's naming that's the hardest)

C47D:
Maybe `Z_TOUCH_DETECTION_THRESHOLD`, [here](https://github.com/lvgl/lv_port_esp32/pull/209) I exposed the z threshold in the menuconfig, and added a function so the user can change the threshold at runtime, but maybe that's an overkill.

liebman:
@C47D I've refactored it a bit. One more change that may help is to make the `Kconfig` part more of a choose one of three.

- IRQ only (default for backwards compatibility)
- IRQ & Touch
- Touch only

Thoughts?

liebman:
Tried the `Kconfig` choice, let me know what you think.

C47D:
Hi, it looks good, I don't have any hardware to test it, is it ready to merge?

liebman:
Yes :-)

C47D:
Thanks for the work you've put into this. I will try to test this with hardware once I get some free time.

PR #23, opened 2021-01-20
Feat/ssd1306 partial update

C47D:
I've working on improving the SSD1306 display driver, hopefully this changes also help us to start working on the MCU independent drivers.

TO DO:
- [x] Update comments
- [x] Handle errors on data transmission


lvgl_tft/ssd1306.c
```diff
@@ -153,96 +148,92 @@ void ssd1306_set_px_cb(lv_disp_drv_t * disp_drv, uint8_t * buf, lv_coord_t buf_w
...
+        0x00,
+        OLED_CMD_SET_COLUMN_RANGE,
+        (uint8_t) area->x1,
+        (uint8_t) area->x2,
+        OLED_CMD_SET_PAGE_RANGE,
+        row1,
+        row2,
+    };
+
+    send_bytes(disp_drv, conf, sizeof conf);
```
kisvegabor, line 167: Is it valid this way? Shouldn't be `sizeof(conf)`?

kisvegabor (commented):
Looks good, just a small question.

lvgl_tft/ssd1306.c
embeddedt: `sizeof` is valid without parentheses if used on expressions, according to [Stack Overflow](https://stackoverflow.com/a/28075117).

C47D: I can add the parentesi can add the parentheses, according to the link they doesn't hurt.

C47D:
This PR also change the display buffer size, from 8192 to 1024 bytes when using the 128*64 px display. As far as I know, the display buffer for monochrome displays should be "in pixels", am I right?

lvgl_tft/ssd1306.c
C47D: Fixed in 9cdc350

kisvegabor: Thanks! :)

kisvegabor:
> As far as I know, the display buffer for monochrome displays should be "in pixels", am I right?

Yes, it's in pixels. So 128 x 64 / 8 = 1 kB is correct.

C47D:
This is ready to be reviewed, I choosed assert over error handling on data transmission because it shouldn't fail.

embeddedt (approved):

PR #24, opened 2021-01-21
Add drivers configuration file

C47D:
This PR is intended to enable driver using a header file, we will also need to update the Kconfig files, so the users can enable drivers by the menuconfig tool. The `CMakeLists.txt` file will also need to be updated.

Any suggestions @kisvegabor @embeddedt ?

### DIFF
--- a/lv_esp32_drivers_conf.h
+++ b/lv_esp32_drivers_conf.h
@@ -1,0 +1,31 @@
+#ifndef LV_ESP32_DRIVERS_CONF_H
+#define LV_ESP32_DRIVERS_CONF_H
+
+/* Enable display drivers */
+#define LV_DISPLAY_EVE      0
+#define LV_DISPLAY_FT81X    0
+#define LV_DISPLAY_GC9A01   0
+#define LV_DISPLAY_HX8357   0
+#define LV_DISPLAY_IL3820   0
+#define LV_DISPLAY_ILI9341  0
+#define LV_DISPLAY_ILI9481  0
+#define LV_DISPLAY_ILI9486  0
+#define LV_DISPLAY_ILI9488  0
+#define LV_DISPLAY_JD79653A 0
+#define LV_DISPLAY_RA8875   0
+#define LV_DISPLAY_SH1107   0
+#define LV_DISPLAY_SSD1306  0
+#define LV_DISPLAY_ST7735S  0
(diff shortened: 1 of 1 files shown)


embeddedt:
I apologize for my forgetfulness, but the long-term plan is to eventually merge this and `lv_drivers` together, right?

C47D:
That's no problem @embeddedt, the plan for this repo is to have `lv_drivers` as a submodule and provide the necessary code so the drivers can be used on the ESP32.
On the other PR I tried to move all the platform dependent code into one place, so we can work towards generic drivers.

C47D:
I have a proof of concept [here](https://github.com/C47D/psoc_lvgl/blob/3693c753de89fb04469f0dd8ea1cd8b2f215feb3/main.c#L155), where the `user_data` field of `lv_disp_drv_t` is used to pass [user callbacks](https://github.com/C47D/psoc_lvgl/blob/3693c753de89fb04469f0dd8ea1cd8b2f215feb3/driver_generic.h#L26-L33), so the driver can use them, like [here](https://github.com/C47D/psoc_lvgl/blob/3693c753de89fb04469f0dd8ea1cd8b2f215feb3/ili9341/ili9341.c#L125-L137):
```c
void ili9341_reset(lv_disp_drv_t *drv)
{
    driver_generic_t *driver = (driver_generic_t *) drv->user_data;
    
	/* Reset the display */
    if (driver->hw_reset) {
        driver->hw_reset(drv);
    } else {
        /* Software reset when no hw_reset user callback is available */
        ili9341_send_cmd(drv, ILI9341_CMD_SWRESET);
        driver->delay_ms(drv, 5);
    }
}
```

It's just an approach, nothing is decided yet.

kisvegabor:
@C47D 
> I have a proof of concept here, where the user_data field of lv_disp_drv_t is used to pass user callbacks, so the driver can use them, like here:

I like the idea of adding the low level driver/interface callbacks to `user_data`.
I assume in the end you have also imagined somehing like this:
```c
driver_generic_t ili9341_driver = {...};
lv_disp_drv_t disp_drv;
/* `flush_cb`, `rounder_cb`, `set_px_cb`, and `user_data` correctly*/
lv_drv_ili9341_create(&disp_drv, &ili9341_driver, /*other params if required?*/);
```

PR #32, opened 2021-02-08
Feature/esp idf component

tore-espressif:
Hello LVGL maintainers,

in this PR there are two notable changes:

1. `CMakeLists.txt` and `KConfig` files from subfolder `lvgl_tft` and `lvgl_touch` were moved to the root of this component. This way, `idf.py` can automatically find them if this repository is cloned to the `componets` directory in project root. Hence, [this added line](https://github.com/lvgl/lv_port_esp32/blob/master/CMakeLists.txt#L6)  is no longer needed.
2. Preparation for ESP-package manager. IDF package mainfest file `idf-component.yml` is added and GNU Makefiles are removed, as they won't be supported by the package manager.

Last but not least, it fixes a bug on ESP-Wrover-kit where the display was always rendered inverted.
I'd like start a discussion (under a new issue #33 ) about reworking the way LVGL drivers rotate/invert the display.


C47D:
Hi,

Is there any public information about the ESP-package manager? What's the minimum ESP-IDF version it will support?

Is there any way for us to keep the component.mk files? Some users are still using them.

tore-espressif:
Thanks for the quick reply.

The first roll-out is planned for 3/2021. Minimum supported ESP-IDF version is going to be 4.0, but I agree that the .mk file should stay for anyone who is still using them (ESP-IDF 3.x).

I will remove point 2 from this PR and resubmit it when it is in better shape.

What do you think about point no 1?

tore-espressif:
I'll close this PR and start a new one.
(shouldn't have started one PR with 2 changes in the beginning...)

C47D:
@tore-espressif I no longer have the Wrover Kit, so I trust you've tested it. And the CMakeLists.txt and Kconfig modifications looks good to me, have you tried to build the `lv_esp32_port` with those modifications?

tore-espressif:
Yes, I have. I used `lv_esp32_port` as basis for my debugging.

PR #34, opened 2021-02-10
Adds soft reset option to ST7789 driver

cranial-smoke:
Thanks for maintaining this library!

Some hardware, like the TTGO T-Watch 2020 V1, do not use a pin for resetting the controller. This PR adds a soft reset option for the ST7789 driver.


C47D:
Hi,

Thanks for the request, it seems correct to me, I think we could hide the reset pin GPIO selection option from the menuconfig if software reset is used, what do you think?

cranial-smoke:
That sounds like a good idea, I will update the branch as soon as I get a chance.

PR #42, opened 2021-02-23
Support make build system

C47D:
@tore-espressif I wasn't sure if those checks needs to be on the lvgl_tft component.mk file or in the "main" one at the root directory.

PR #43, opened 2021-02-23
Feature/refactor build systems

tore-espressif:
Hello @C47D 

- This PR tries to define everything that is needed for correct compilation of this repository in both GNU Make and idf.py(CMake).
- Now it is a single IDF component, so the user does not have to define `LV_LVGL_H_INCLUDE_SIMPLE` and `EXTRA_COMPONENTS_DIR` manually, which makes it a lot easier for first-time user

ATM it is not possible to compile `lv_port_esp32` with this version of drivers with Make. The LVGL has to be updated too:

this line: https://github.com/lvgl/lvgl/blob/master/component.mk#L3 resolves to `false` and IDF treats it as empty `component.mk` which creates mess. Have you tried building `lv_port_esp32` with GNU Make? Maybe the problem is only on my machine...


C47D:
> ATM it is not possible to compile lv_port_esp32 with this version of drivers with Make. The LVGL has to be updated too:
this line: https://github.com/lvgl/lvgl/blob/master/component.mk#L3 resolves to false and IDF treats it as empty component.mk which creates mess. Have you tried building lv_port_esp32 with GNU Make? Maybe the problem is only on my machine...

Yes, I've tried and also got a mess.

I've opened this https://github.com/lvgl/lv_port_esp32/pull/262 to update the project main Makefile first, can you please take a look at it?

C47D:
See https://github.com/lvgl/lvgl_esp32_drivers/pull/42#issuecomment-784315050

tore-espressif:
After we merge this, https://github.com/lvgl/lv_port_esp32/pull/262/commits/b2553d03754ff4d01d58f86d055ef04904c47844 will not be needed any more. 

My proposal:
1. Merge #43 (my branch to your branch)
2. Merge #42  (your branch to master)
3. Close lvgl/lv_port_esp32#262 and I'll prepare PR in lvgl/lv_port_esp32 that updates submodule `lvgl_esp32_drivers'`to master and cleans CMakeLists.txt and component.mk files

Then the only remaining thing is lvgl/lvgl/component.mk file and it should all work with both CMake and GNU Make.

Issue #33, opened 2021-02-09
ILI9341 inversion on ESP32-Wrover-Kit


tore-espressif:
Hello,

ILI9341 display on ESP32-Wrover-Kit is always inverted, regardless off the settings `Display orientation` in menuconfig.
Problem lays on [line 232](https://github.com/lvgl/lvgl_esp32_drivers/blob/master/lvgl_tft/ili9341.c#L232) where correctly it should be
``` C
// 6C - portrait
// EC - portrait inverted
// CC - landscape
// 4C - landscape inverted
uint8_t data[] = {0x6C, 0xEC, 0xCC, 0x4C};
```

Nevertheless, I don't consider this solution satisfactory. This is screenshot from [ILI9341 datasheet](https://cdn-shop.adafruit.com/datasheets/ILI9341.pdf) 

![image](https://user-images.githubusercontent.com/77206109/107367249-60d8bc00-6adf-11eb-8206-a1b795c47634.png)

There are 3 bits relevant for the display orientation and axis inversion: `MY, MX, MV` resulting in 8 different combinations.

My proposal is to split the `Display orientation` settings into three:

1. Display orientation -> Landscape/portrait
2. Invert X axis -> true/false
3. Invert Y axis -> true/false

Then this [section of code](https://github.com/lvgl/lvgl_esp32_drivers/blob/master/lvgl_tft/ili9341.c#L229-235) and similar in other drivers can be removed and the default settings for each board could be resolved by KConfig. (e.g. selecting Wrover-kit would set Invert X: true and Invert Y: false)

Should be handled together with #3 and #21, any opinion is welcome.

Thanks

---
**NOTE**

Not to be mistaken with color inversion. (command 0x20 and 0x21). #38 

---

PR #46, opened 2021-03-02
Fix/wrover kit display inversion

tore-espressif:
Correct display inversion of ESP32-Wrover-Kit

hot-fix for #33


C47D:
Hi @tore-espressif is this still up to date?

tore-espressif:
Yes, it is. This is correct configuration for Wrover kit.

PR #58, opened 2021-04-28
GPIO display reset optional / Added M5Core2 / White space fixes

ropg:
Three separate commits in one PR:

&nbsp;

**`LV_DISP_USE_RST`**

This is the same fix as @usedbytes has made [here](https://github.com/usedbytes/lvgl_esp32_drivers) before, making the use of a GPIO pin for reset optional. Some people have the reset tied, or have it hooked up to a pin of a power management IC. No need to waste a GPIO.

@usedbytes's fix is only implemented the Kconfig option in the IL9341 driver, this one is implemented in all drivers where there's a GPIO for reset.

&nbsp;

**`LV_PREDEFINED_DISPLAY_M5CORE2`**

Added support for the M5Core2. This depends on the previous, because it is setting `CONFIG_LV_DISP_USE_RST=n`

&nbsp;

**White space**

I used a separate commit to remove white space at EOL and added some EOLs at EOF where needed. People's code editors are often set to do this automatically and it can make commits where they don't belong hard to audit. Also added a few entries to .gitignore.


kisvegabor:
Thanks. Although I'm not deep into this repo but the changes look good so I merge it.

PR #69, opened 2021-06-11
Feature/esp32c3 support

C47D:
@dastarling Could you take a look at this and test it when you have time?

Fixes https://github.com/lvgl/lv_port_esp32/issues/269


lvgl_touch/touch_driver.c
```diff
@@ -40,6 +42,8 @@ bool touch_driver_read(lv_indev_drv_t *drv, lv_indev_data_t *data)
     res = FT81x_read(drv, data);
 #elif defined (CONFIG_LV_TOUCH_CONTROLLER_RA8875)
     res = ra8875_touch_read(drv, data);
+#elif defined (CONFIG_LV_TOUCH_CONTROLLER_GT911)
+    res = gt911_read(drv, data);
 #endif
 
     return res;
```
embeddedt, line 49: Note that this needs an update for v8 as the return value is ignored there.

You can do it without adjusting the drivers with this change:

```suggestion
#if LVGL_VERSION_MAJOR < 8
    data->continue_reading = res;
#else
    return res;
#endif
```

You would also have to conditionally change the prototype of `touch_driver_read` to return `void`.

C47D: Thanks for the information @embeddedt , I think the `#if LVGL_VERSION_MAJOR < 8` must be `#if LVGL_VERSION_MAJOR >= 8` instead, isn't?

embeddedt: Yes; I mixed up the two versions. Sorry.

I think you only need to update `touch_driver.c` so you can probably keep that change in this PR.

lvgl_touch/touch_driver.h
```diff
@@ -44,7 +44,12 @@ extern "C" {
  * GLOBAL PROTOTYPES
  **********************/
 void touch_driver_init(void);
+
+#if LVGL_VERSION_MAJOR >= 8
```
C47D, line 48: @embeddedt is this change what you meant?

lvgl_touch/touch_driver.c
C47D: @embeddedt is this change what you meant?

embeddedt: Yes; both changes are perfect.

embeddedt: Yes; both changes are perfect.

C47D:
@kisvegabor touch_driver got an update for LVGLv8, do you think we need any other update for that for this particular PR?

lvgl_helpers.c
```diff
@@ -19,9 +19,17 @@
 #include "driver/i2c.h"
 
 #ifdef LV_LVGL_H_INCLUDE_SIMPLE
-#include "src/lv_core/lv_refr.h"
+#if (LVGL_VERSION_MAJOR < 8)
```
kisvegabor, line 22: For the sake of simplicity, I think we can include `lvgl.h` directly.

lvgl_touch/gt911.c
```diff
@@ -0,0 +1,168 @@
...
+    }
+}
+
+/**
+  * @brief  Get the touch screen X and Y positions values. Ignores multi touch
+  * @param  drv:
+  * @param  data: Store data here
+  * @retval Always false
+  */
+bool gt911_read(lv_indev_drv_t *drv, lv_indev_data_t *data) {
```
kisvegabor, line 117: Should return `void`.

kisvegabor (commented):
Looks good in general, only a few comments above. 

On a general note, for v8 instead of adding `#if LVGL_VERSION_MAJOR >= 8` I started to keep only v8 stuff because people who use v7 can use the release/v7 branch.

I'm not sure how well could it work here as the goal is to have as many drivers for as many releases as possible. So if someone adds a driver to v8 maybe the preprocessor instructions are the best way to backport them to v7.  Anyway, I just wanted to mention my new approach in the hope it can help here in some form :slightly_smiling_face:

lvgl_touch/gt911.c
embeddedt: I don't think it's needed; the translation to the v8 approach is done [in `touch_driver_read`](https://github.com/lvgl/lvgl_esp32_drivers/pull/69/files#diff-ecabe01db47fce3a7d73d2e391e5ade59d1560f403097408668c9d2317e976c1R49-R56).

kisvegabor: Does it mean the indev driver in this repo will keep following v7 API?

embeddedt: The top level `read_cb` function (`touch_driver_read`) uses the v8 API, however, to avoid needing to refactor all the individual drivers, they are still using the v7 API.

C47D:
> On a general note, for v8 instead of adding #if LVGL_VERSION_MAJOR >= 8 I started to keep only v8 stuff because people who use v7 can use the release/v7 branch.
I'm not sure how well could it work here as the goal is to have as many drivers for as many releases as possible. So if someone adds a driver to v8 maybe the preprocessor instructions are the best way to backport them to v7. Anyway, I just wanted to mention my new approach in the hope it can help here in some form 🙂

I see, I also wanted to keep a branch for drivers v7 and another for v8, but I don't know how much effort would it take to keep both in sync, as far as I've seen the differences are the indev return value of the read function is now `void` and the rotation support.

embeddedt:
Rotation is also available in 7.11.

lvgl_touch/gt911.c
C47D: What about defining a symbol like `INDEV_READ_RETURN_TYPE` and use it like this `INDEV_READ_RETURN_TYPE gt911_read(lv_indev_drv_t *drv, lv_indev_data_t *data)`. We could define it to `void` when using v8, and `bool` when using v7.

lvgl_helpers.c
C47D: Thanks, I will update it.

C47D:
> Rotation is also available in 7.11.

So the return type of indev read is the only change? Drivers wise.

lvgl_helpers.c
C47D: Like this?
```diff
diff --git a/lvgl_helpers.c b/lvgl_helpers.c
index 56c9183..a215c43 100644
--- a/lvgl_helpers.c
+++ b/lvgl_helpers.c
@@ -19,17 +19,9 @@
 #include "driver/i2c.h"

 #ifdef LV_LVGL_H_INCLUDE_SIMPLE
-#if (LVGL_VERSION_MAJOR < 8)
-    #include "src/lv_core/lv_refr.h"
-#else
-    #include "src/core/lv_refr.h"
-#endif
+#include "lvgl.h"
 #else
-#if (LVGL_VERSION_MAJOR < 8)
-    #include "lvgl/src/lv_core/lv_refr.h"
-#else
-    #include "lvgl/src/core/lv_refr.h"
-#endif
+#include "lvgl/lvgl.h"
 #endif
```

embeddedt:
I believe that and the requirement for drivers to be `static` are the only breaking changes; yes.

kisvegabor:
> I believe that and the requirement for drivers to be static are the only breaking changes; yes.

And in display driver there are some renames: `disp_buf` -> `draw_buf`.

lvgl_touch/gt911.c
kisvegabor: > What about defining a symbol like INDEV_READ_RETURN_TYPE and use it like this INDEV_READ_RETURN_TYPE gt911_read(lv_indev_drv_t *drv, lv_indev_data_t *data). We could define it to void when using v8, and bool when using v7.

This way the last `return false;` still needs to be `#if`ed. 

I think it's not a problem if the `bool` return value is kept if it makes the handling of v7 easier. I just wanted to be sure it's intentional. :slightly_smiling_face:

C47D:
Ok, will do the initial changes for v8 on this driver, some of the changes are necessary in `lv_port_esp32` instead.

C47D:
I guess this driver is complete, waiting for @dastarling for feedback, will work on v8 support for display drivers on another PR

dastarling:
Thanks for all the work.  I'll give it a check soon.  I think I am still on v7, was thinking of updating to v8, but there were some differences I wasn't prepared to go thru at the time.  I'll try to test it at v7 and see if I can then update to v8 and test again.

C47D:
> Thanks for all the work. I'll give it a check soon. I think I am still on v7, was thinking of updating to v8, but there were some differences I wasn't prepared to go thru at the time. I'll try to test it at v7 and see if I can then update to v8 and test again.

v7 is fine, support for v8 will take some work. Thanks for the help :), there's also a new PR to handle better I2C (#70 ), but it's not merged just yet.

C47D:
@kisvegabor @embeddedt Think we need to merge this, some new PRs are replicating some fixes available here, what do you think?

kisvegabor (approved):

kisvegabor:
Sure!

dastarling:
I tested Master and it builds and runs with lvgl v7.11.0.  But the ili9488.c file does not include some changes I made to the flush function due to the limited memory of the C3.  So the graphics output is banded on the screen with Resolution 480x320.
With lvgl v8.01 I get these compile errors:
`./components/lvgl_esp32_drivers/lvgl_helpers.h:54:25: error: 'LV_HOR_RES_MAX' undeclared (first use in this function); did you mean 'LV_HOR_RES'?
 #define DISP_BUF_SIZE  (LV_HOR_RES_MAX * 40)
                         ^~~~~~~~~~~~~~
../main/main.c:95:41: note: in expansion of macro 'DISP_BUF_SIZE'
     lv_color_t* buf1 = heap_caps_malloc(DISP_BUF_SIZE * sizeof(lv_color_t), MALLOC_CAP_DMA);
                                         ^~~~~~~~~~~~~
../components/lvgl_esp32_drivers/lvgl_helpers.h:54:25: note: each undeclared identifier is reported only once for each function it appears in
 #define DISP_BUF_SIZE  (LV_HOR_RES_MAX * 40)
                         ^~~~~~~~~~~~~~
../main/main.c:95:41: note: in expansion of macro 'DISP_BUF_SIZE'
     lv_color_t* buf1 = heap_caps_malloc(DISP_BUF_SIZE * sizeof(lv_color_t), MALLOC_CAP_DMA);
                                         ^~~~~~~~~~~~~
../main/main.c:107:12: error: unknown type name 'lv_disp_buf_t'
     static lv_disp_buf_t disp_buf;
            ^~~~~~~~~~~~~
../main/main.c:122:5: error: implicit declaration of function 'lv_disp_buf_init'; did you mean 'lv_disp_drv_init'? [-Werror=implicit-function-declaration]
     lv_disp_buf_init(&disp_buf, buf1, buf2, size_in_px);
     ^~~~~~~~~~~~~~~~
     lv_disp_drv_init
../main/main.c:128:13: error: 'lv_disp_drv_t' {aka 'struct _lv_disp_drv_t'} has no member named 'buffer'
     disp_drv.buffer = &disp_buf;
             ^
cc1: some warnings being treated as errors
[1073/1227] Building C object esp-idf/main/CMakeFiles/__idf_main.dir/gui_sm.c.obj
../main/gui_sm.c: In function 'main_menu':
../main/gui_sm.c:196:85: warning: passing argument 2 of 'lv_obj_get_child' makes integer from pointer without a cast [-Wint-conversion]
         current_sensor_name = lv_label_get_text(lv_obj_get_child((lv_obj_t *)param, NULL));
                                                                                     ^~~~
In file included from ../components/lvgl/src/core/lv_obj.h:121,
                 from ../components/lvgl/lvgl.h:32,
                 from ../main/gui_sm.c:22:
../components/lvgl/src/core/lv_obj_tree.h:136:20: note: expected 'int32_t' {aka 'long int'} but argument is of type 'void *'
 struct _lv_obj_t * lv_obj_get_child(const struct _lv_obj_t * obj, int32_t id);
                    ^~~~~~~~~~~~~~~~
`

dastarling:
The ESP32C3 displays this message in the terminal indicating the lack of memory for a full screen DMA transfer at the 480x320 resolution

> W (157073) ILI9488: Could not allocate enough DMA memory!

dastarling:
Should I open a separate case for the lack of memory?

dastarling:
I also found a few other minor issues;

lvgl_esp32_drivers\lvgl_tft\Kconfig: 828
range 0 21 if IDF_TARGET_ESP32C3 -> range **-1** 21 if IDF_TARGET_ESP32C3

lvgl_esp32_drivers\lvgl_tft\Kconfig: 840
range 0 21 if IDF_TARGET_ESP32C3 -> range **-1** 21 if IDF_TARGET_ESP32C3

lvgl_esp32_drivers\lvgl_touch\Kconfig:139
-> **range 0 21 if IDF_TARGET_ESP32C3**

Issue #72, opened 2021-06-17
Design proposal


tore-espressif:
# Refactoring proposal

Here is a  summary of major/refactoring issues that must be tackled systematically in order to make this SW component maintainable and extandable:
## Issues collection

1. Previous refactoring plans:
    * Kconfig refactoring https://github.com/lvgl/lvgl_esp32_drivers/issues/40
    * Long-term plans https://github.com/lvgl/lvgl_esp32_drivers/issues/44
2. LVGL v8 https://github.com/lvgl/lvgl_esp32_drivers/issues/68
3. Versioning https://github.com/lvgl/lv_port_esp32/issues/272
4. Multiple displays
https://github.com/lvgl/lvgl_esp32_drivers/issues/31
https://github.com/lvgl/lv_port_esp32/issues/276
https://github.com/lvgl/lv_port_esp32/issues/277
5. New graphical interfaces (RGB, parallel)
https://github.com/lvgl/lvgl_esp32_drivers/issues/64
6. Runtime configuration and Kconfig
https://github.com/lvgl/lvgl_esp32_drivers/issues/65
https://github.com/lvgl/lv_port_esp32/issues/282
7. New Espressif devices
https://github.com/lvgl/lv_port_esp32/issues/269

## Proposal:

There are many hints in the issues' discussions that imply that the repo needs a major refactoring, so I'd like to present my high-level view of what and how should be done.
Please don't take it as high-jacking or critique of your wonderful project, but rather as a willingness to support it :)

### 1. Abandon Kconfig
Nice article about [What not to turn into Kconfig option](https://docs.zephyrproject.org/2.3.0/guides/kconfig/tips.html#what-not-to-turn-into-kconfig-options) in Zephyr's project documentation, most notably __pin numbers__.

In case of ESP-LVGL drivers it means __all__ the options.

Many issues from users are about Kconfig configuration, and at the same time it gives us no benefit.
On the contrary, it complicates our effort on any new feature, as it must be reflected in Kconfig.

Getting rid of Kconfig over runtime configuration would help us tackle issues in group 1, 4, 5, 6, 7 __on top of a good design.__

#### 1.1. Eval board support
Support for evaluation boards (Wrover-kit, M5Stack...) is currently done only through Kconfig.
As a replacement, examples of how to configure the drivers and LVGL will be provided in examples folder (see point 4).

### 2. Use `esp_lcd` component
[...]

PR #74, opened 2021-06-20
st7735s: Added brightness control though LED pin

Rajssss:
-> Uses LED(Backlight) Pin as PWM to control the brightness

Signed-off-by: Rajssss <sssraj.sssraj@gmail.com>


tore-espressif:
Hello @Rajssss 
Thank you very much for you PR ! Backlight control is something that is needed by many users.

However, I don't like adding this feature to a specific display driver (ST7735S), because it is not specific to this driver, but rather common for (almost) all displays.

I'd like to carve-out the backlight control from our current code and create a component, that could be as simple as the code you pushed. But reusable for all displays.

Let me know what you think

Rajssss:
> Hello @Rajssss
> Thank you very much for you PR ! Backlight control is something that is needed by many users.
> 
> However, I don't like adding this feature to a specific display driver (ST7735S), because it is not specific to this driver, but rather common for (almost) all displays.
> 
> I'd like to carve-out the backlight control from our current code and create a component, that could be as simple as the code you pushed. But reusable for all displays.
> 
> Let me know what you think

I agree, that will be the best, but I don't know about all other controllers supported by this project. But, If it's all about a backlight pin then it can be done, though right now I don't have any idea of a separate component. I will look into it, but not sure how much time it will take.

tore-espressif:
Best and future-proof solution would be to create new component `esp_lcd_backlight.c/.h`. So if you are willing to create it, that would be great.

If not, we can merge it as it is, and refactor the code in upcoming 2.0.0 refactoring #72

Rajssss:
> Best and future-proof solution would be to create new component `esp_lcd_backlight.c/.h`. So if you are willing to create it, that would be great.
> 
> If not, we can merge it as it is, and refactor the code in upcoming 2.0.0 refactoring #72

Ok, I will try to do that by this weekend.

Rajssss:
@tore-espressif please review it, I have not edited the kconfig for now, is it required?

C47D:
It seems to me it's going in the right direction, thanks @Rajssss, @tore-espressif do you think we need to add a `choice` with the different back-light control options? We could have:
- no back-light control
- on/off back-light control
- pwm back-light control (this one)

Rajssss:
> It seems to me it's going in the right direction, thanks @Rajssss, @tore-espressif do you think we need to add a `choice` with the different back-light control options? We could have:
> 
>     * no back-light control
> 
>     * on/off back-light control
> 
>     * pwm back-light control (this one)

I like the idea, but in a long run maybe, as we have to test different displays, maybe.

C47D:
@Rajssss Thanks :)

tore-espressif:
LGTM!

Rajssss:
> The only missing thing seems to be the Kconfig. Apart from that, LGTM!
> 
> > I like the idea, but in a long run maybe, I will see how much I can do this weekend
> 
> Understand. We can merge it with pwm only now and fine tune it later.

Thanks :)

C47D:
Sorry for the late reply 😊

tore-espressif:
Oh, sorry, I forgot about this PR. We still need to add Kconfig option CONFIG_LV_ENABLE_BACKLIGHT_CONTROL.

I'll add it today

PR #92, opened 2021-07-30
Display hardware abstraction for IOs

C47D:
This is an idea of how can we start removing the ESP32 independent code from the display drivers. In this example I've used the ST7789 driver. The `display_bsp` header and source files are the ones that keep all the MCU related code.


Rajssss:
Looks like a good idea, but I have two questions.
Isn't this repo meant for esp32 display drivers only? So why such abstraction is needed?
Also is this because, in future this repo will be used as a submodule with lv_drivers?

C47D:
> Also is this because, in future this repo will be used as a submodule with lv_drivers?

We're planning to move the "MCU independent" drivers into the `lv_drivers` repo, used it as submodule in here and have the ESP32 code separated.

C47D:
Any thoughts about this @kisvegabor @embeddedt @tore-espressif ?

kisvegabor:
`driver/gpio.h` is not existing yet, right?

C47D:
Yes it does, code on the `lvgl_tft/display_bsp.c` file is ESP (or any other MCU) specific. Maybe renaming `lvgl_tft/display_bsp.c` to `lvgl_tft/display_esp_bsp.c` would be better? Another MCU source file would be named `lvgl_tft/display_stm32_bsp.c` for example.

Now that I think about it `bsp` (Board Support Package) doesn't sound right to me, does `hal` (Hardware Abstraction Layer) fits better the purpose of those files?

kisvegabor:
>  Maybe renaming lvgl_tft/display_bsp.c to lvgl_tft/display_esp_bsp.c would be better?

I agree that it'd be better to call create vendor specific hal files. Do we need separate display, indev and tick files?

C47D:
> I agree that it'd be better to call create vendor specific hal files. Do we need separate display, indev and tick files?

I tend to separate things, but maybe having just one file named `lvgl_MCU_bsp.c` for all of those three would be better, what do you think?

kisvegabor:
I think there would be many redundant content. E.g. I2C can be used for indev and display too. So it might be simpler to handle them in one place. 

I can imagine these functions: (The name are not suggestions, just for illustration).

 - init()
 - delay()
 - SPI
   - spi_clock_speed_set()
   - spi_dc_set()
   - spi_cs_set()
   - spi_read()
   - spi_send_data()
   - spi_send_color()
 - I2C
   - i2c_read()
   - i2c_send_data()
   - i2c_send_color()
   - i2c_clock_speed_set()
 - Parallel 
   - parallel_set_clock_speed()
   - parallel_send_data()
   - parallel_send_color()
 - Framebuffer 
   - framebuffer_write()
 - MIPI
   - mipi_send_data()
   - mipi_send_color()  

The first parameter can be the ID of the periphery. E.g.
```c
#define DISP_SPI_ID SPI3

spi_send_data(DISP_SPI, buf, 32);
```

The functions ending in `send_color()` should call `lv_disp_flush_ready()` too as needed.

What do you think?

C47D:
In my opinion setting up the peripherals should be done by the user, even before calling any LVGL code, sometimes we use tools to autogenerate configuration code, such as STM32CubeMX or Atmel Start, so I would remove any `x_clock_speed_set` functions.

> The functions ending in send_color() should call lv_disp_flush_ready() too as needed.

That's not possible in some drivers, the e-ink ones can't do that as they are right now.

I will check the i2c manager code to see how is it implemented.

kisvegabor:
> In my opinion setting up the peripherals should be done by the user, even before calling any LVGL code, because sometimes we use tools to generate the configuration code, such as STM32CubeMX or Atmel Start, so I would remove any x_clock_speed_set functions.

Sounds reasonable.

C47D:
I will add the fixes to this PR and see how easy or difficult it goes.

C47D:
@tore-espressif Thanks for marking it as draft, I couldn't change it myself.

> In the lvgl repo we use lv_ perfix, so for the sake of consistency I suggest lv_port_*.

Thanks for the suggestion, I'm pretty bad at giving names, I will update the code with this 😄 

> What is the relation between display_hal_backlight and st7789_enable_backlight?

Inside `st7789_enable_backlight` we call `display_hal_backlight`, passing a pointer to the user `lv_disp_drv_t`, so if the user have multiple displays we could do:

```c
void display_hal_backlight(lv_disp_drv_t *drv, uint8_t state)
{
    uint32_t pin = 0;

    if (/* Check if drv points to st7789 driver */) {
        pin = st7789_backlight_pin;
    }
    else if (/* Check if drv points to the other driver */) {
        pin = other_display_backlight_pin;
    }
    else {
        assert(0); /* We shouldn't reach here */
    }

    gpio_set_level(pin, state);
}
```

So all the ESP32 code is moved from the display drivers into the `lv_port_esp32` file.

> Shouldn't we have display_hal_orientation too if we have st7789_set_orientation?

Yes, I'm still thinking on how to do this 😅 

Just a brain dump so I don't forget this:
I was thinking on how could we abstract the different peripherals that some display controllers support, e.g. ILIx displays support SPI and parallel. So maybe the easiest way to do it is something like this:

```c
void ili_send_data(void *data, size_t length)
{
#if defined ILI_TRANSPORT_SPI
    display_hal_spi_send(data, length);
#elif defined ILI_TRANSPORT_PARALLEL
    display_hal_parallel_send(data, length);
#elif defined ILI_TRANSPORT_CUSTOM
    display_hal_custom_send(data, length);
#else
#error You need to select one transport.
#endif
}
```

kisvegabor:
> Thanks for the suggestion, I'm pretty bad at giving names, I will update the code with this smile

People [say](https://martinfowler.com/bliki/TwoHardThings.html) it's complicated: 
image
:laughing: 

> I was thinking on how could we abstract the different peripherals that some display controllers support, e.g. ILIx displays support SPI and parallel. So maybe the easiest way to do it is something like this:

Looks good. I know that it was just a quick example but `drv` should be part of these functions too. right?

Could you comment on my last question too?
> What is the flow of the initialization?
> static lv_disp_drv_t disp_drv;
lv_disp_drv_init(&disp_drv);
disp_driver_init(&disp_drv);
// Do I need to assign flush_cb or will it be done in disp_driver_init -> st7789_init? If it needs to be assigned manually what is point of abstracting only the init function?
// If I also wan't to pass resolution of the init function how can I do it?

For the libinput driver (handles keyboard, touch, mouse, etc on Linux) it'd be nice to automatically add all the available indevs as lvgl drivers. To do this an `lv_libinput_init()` function would be re required that allocates and initializes all the drivers. However I'm not sure how it fits into the current schema and what could be the possible drawback in case of simpler embedded configuration (e.g. only one touchpad). (Related PR and comment: https://github.com/lvgl/lv_drivers/issues/151#issuecomment-913202619)

C47D:
So I had to rebase master into this branch (it was about 66 commits behind), and then forced a push into this remote branch, I guess it would be better not to merge this PR but rather create a new one once the proof of concept is working as expected. The commit history would be a lot cleaner as well.

I've been testing this PR (before the rebase) on hardware but I didn't got it working, which seems a bit odd since the changes only were moving IO operations to another file, is really nice to have the github actions setup up 😄

C47D:
So this is ready for review, but it does have some breaking changes:

- Display is not initialized when calling `lvgl_driver_init`, it should be renamed to `init_peripheral` or something along those lines because it's what it does.
- Display initialization function, `st7789_init` now takes one argument, a pointer to `lv_disp_drv_t`.

Future work:

- Add configuration header file so we can remove all esp-idf code from `st7789.h`, do something similar to `lv_conf_internal.h`, so we can used Kconfig or not.
- Add pointer to `lcd_init_cmd_t` as argument to `st7789_init` so users can provide custom display configuration, if `NULL`, then we use the default one, the one we currently have.
- Improve `CMakeLists.txt` file.
- Improve Kconfig file, the ST7789 display currently have an option which allows the user to reset the display controller with software, we could fallback to this when a reset pin is not enabled.

C47D:
So I've added a template for the `disp_cb` callback so we could use it to update the display orientation at runtime, not implemented not tested in hardware. I've also moved the ST7789 display configuration to a separated header file, so we don't include any ESP-IDF code directly into the display code.

kisvegabor:
Two architectural related questions are not clear to me. Now there is [display_hal_backlight()](https://github.com/lvgl/lvgl_esp32_drivers/pull/92/files#diff-4ed969e112572f9a3585780631e75703aa656cf480a815ac84af5e19f0a4d896R16) in `lv_port/lv_port_display_espressif.c`.

What if the user has multiple displays, or the backlight needs to be set in trickier way. If the user needs to change it for any reasons he needs to modify the source code of the driver. It's the same for e.g. `display_hal_gpio_rst()`. The reset pins are different for each display so for multiple displays this functions needs to be modified. 

So wouldn't be better to leave these functions for the user too, similarly to `disp_spi_send_colors`? As @tore-espressif suggested earlier we can have examples for some basic configs but in the name of greater flexibility it might be useful to give more freedom for the user. 

Is there a place to see the list of functions that should be implemented by the user (or can you list them)?

The other thing is now [disp_driver_init](https://github.com/lvgl/lvgl_esp32_drivers/pull/92/files#diff-8017a67c23d46b1d9c8c9f407a88b42e0f2341a4acd769c1e34943dfdb075b8eR10) allows using only one display. 

If I see it correctly, now the flow is that you initialize a display driver with `disp_driver_flush`, the resolution, etc and then pass that driver's pointer to `disp_driver_init()`. 

I wonder if we can wrap thes into the init function. My idea is to remove `disp_driver_init` and make the user call `st7789_init` directly with `hor_res`, `ver_res`, and an other parameters. `st7789_init` could `malloc` a driver, set the hor/ver res, flush_cb, set_px_cb if required, register a display, and return the new display. The prototype of `<display>_init` can be slightly different based on the features of the given controller/hardware.  The user has a larger freedom on how the draw buffers are allocated so they also could be passed to the init function. All together: 
```c
lv_disp_t * disp1 = st7789_init(320, 240, buf1, buf2, buf_size, some_flags);
```

Minor thing but setting the backlight in 0..255 range would allow using PWM or so too. We can have defines like `BACKLIGHT_ON/OFF` for 0 and 255. 

What do you think?

---

Regardless to my comments above we can merge this PR to have a cleaner base for further discussions.

C47D:
> What if the user has multiple displays, or the backlight needs to be set in trickier way. If the user needs to change it for any reasons he needs to modify the source code of the driver. It's the same for e.g. display_hal_gpio_rst(). The reset pins are different for each display so for multiple displays this functions needs to be modified.

So the reason to move some of the code to an specific source file (`lv_port_display_espressif.c` in this case) is that the API definitions of the HAL layer will be in just that source file, we are providing only the basic functionality, as we have already discussed, it's impossible to provide a solution for every use case.

So, if the user needs custom behavior they only need to exclude the `lv_port_display_espressif.c` file from their build system, and provide a custom source file that implement those functions, so the display controllers will call their custom implementation.

Here's how I think they handle the reset of multiple displays:

On their main.c they define and configure the different `lv_disp_drv_t`:
```c
/* Global objects */
lv_disp_drv_t front_display;
lv_disp_drv_t left_display;

void main(void)
{

/* The usual configuration in here */
front_display.flush_cb = st7789_flush;
left_display.flush_cb = ssd1306_flush;

/* display registration */
lv_disp_drv_register(&front_display);
lv_disp_drv_register(&left_display);

/* more code... */

}
```

On their custom port file:
```c
void display_hal_gpio_rst(lv_disp_drv_t *drv)
{
  if (front_display == drv) {
    /* Reset front display */
  }
  else if (left_display == drv) {
    /* Reset left display */
  }
  else {
    /* invalid display */
  }
}
```

Same applies for backlight control.

> Is there a place to see the list of functions that should be implemented by the user (or can you list them)?

So far the functions declared in `display_hal.h`, I've only updated the ST7789 display tho, we might need more when updating other displays. I'm actually working on trying to figure out how to provide display rotation at runtime.

> The other thing is now disp_driver_init allows using only one display.

If I see it correctly, now the flow is that you initialize a display driver with disp_driver_flush, the resolution, etc and then pass that driver's pointer to disp_driver_init().

I wonder if we can wrap thes into the init function. My idea is to remove disp_driver_init and make the user call st7789_init directly with hor_res, ver_res, and an other parameters.

I agree, those functions only allow us to use one driver at the project, we need to remove them, they were just helpers.
I was thinking on adding another parameter to the init functions so the user could provide a custom initialization sequence, we keep the basic ones and use them if the user doesn't need anything special.

> Minor thing but setting the backlight in 0..255 range would allow using PWM or so too. We can have defines like BACKLIGHT_ON/OFF for 0 and 255.

@tore-espressif and a user implemented an improved backlight support, but I haven't tested it, it allows PWM control.

kisvegabor:
> So, if the user needs custom behavior they only need to exclude the lv_port_display_espressif.c file from their build system, and provide a custom source file that implement those functions (I think this is called link time substitution), so the display controllers will call their custom implementation.

Got it, thanks. It'd be even better if the user could enable/disable configs to enable/disable certain ports. If none of them is enabled he needs to add his own port.

> So far the functions declared in display_hal.h, I've only updated the ST7789 display tho, we might need more when updating other displays. I'm actually working on trying to figure out how to provide display rotation at runtime.

`disp_spi_send_colors` should be there as well, right?


> I agree, those functions only allow us to use one driver at the project, we need to remove them, they were just helpers.
I was thinking on adding another parameter to the init functions so the user could provide a custom initialization sequence, we keep the basic ones and use them if the user doesn't need anything special.

IMO having a custom init sequence is a very special case. Instead the driver should support the common settings (color format, orientation, flipping, etc). I'd rather change `st7789_init` like:
```c
st7789_cfg_t cfg;
st7789_cfg_init(&cfg);
cfg.back_porch=12;
cfg.px_clock=10000000;
..etc...

lv_disp_t * disp1 = st7789_init(320, 240, buf1, buf2, buf_size, &cfg);
``` 

If `cfg` is `NULL` go with some default settings.

`cfg` can be different for every display. E.g. an `st7789` is very different from a Linux framebuffer and we can't hope that they can described by the same descriptor. E.g. 
```c
fbdev_cfg_t cfg;
fbdev_cfg_init(&cfg);
fbdev.path = "/dev/fb0";

lv_disp_t * disp1 = fbdev_init(320, 240, buf1, buf2, buf_size, &cfg);

---

We can update attach additional callbacks to the display driver. E.g. a `set_brightness_cb`. Or it could be only a value in `disp_drv` and the `drv_update_cb` can handle it. 

What do you think?

I think using the  `<driver>_init()` functions directly would result in higher flexibility and easier use. The user needs to call only a function, and that's all. He doesn't need to know anything about `flush_cb`, `set_px_cb` or other "magics".

**Comment by C47D** [comment]
> Got it, thanks. It'd be even better if the user could enable/disable configs to enable/disable certain ports. If none of them is enabled he needs to add his own port.

That sounds good, so they can use some of our ports and some of theirs, isn't?

> disp_spi_send_colors should be there as well, right?

Yes, same for I2C.

> IMO having a custom init sequence is a very special case. Instead the driver should support the common settings (color format, orientation, flipping, etc). I'd rather change st7789_init like:

So, the `lv_disp_drv_t` is allocated inside the init function by LVGL?

> We can attach additional callbacks to the display driver. E.g. a set_brightness_cb. Or it could be only a value in disp_drv and the drv_update_cb can handle it.
This way lvgl could provide lv_disp_set_brightness() similarly to lv_disp_set_rotation().
What do you think?

That sounds really useful, I was thinking on using some concepts from #1 to create a custom object with a display controller, backlight control and indev controller.

Maybe @tore-espressif have more ideas, or questions about this PR. What do you think about merging this into development and create new PRs for the improvements you mentioned above?

**Comment by kisvegabor** [comment]
> That sounds good, so they can use some of our ports and some of theirs, isn't?

Yes, e.g. if the the ESP port is good for them as it they can use it, or copy and modify it, or create brand new one.

> So, the lv_disp_drv_t is allocated inside the init function by LVGL?

Not by LVGL but in `st7789_init`.

**Comment by C47D** [comment]
@kisvegabor Found a couple of issues while trying to use the rotated feature on the display:

1. `lv_disp_set/get_rotation` are not documented in the v7.11 version of the docs, I didn't knew where to send the PR to fix it, nor the place to fix it 😓 
2. v7 doesn't currently have a user callback when the display rotation is updated, so I tried to get the display rotation on the flush callback (and update it if needed), but the `lv_disp_get_rotation` function expects a pointer to `lv_disp_t`, is there a way to get the current `lv_disp_t`? I tried doing an upcast, but I think it's not legal 😸 
```c
lv_disp_t *display = (lv_disp_t *) drv;
```

embeddedt:
> I tried doing an upcast, but I think it's not legal

I believe it works in v7 (as the driver is stored at the very beginning of the `lv_disp_t` structure), however, it does *not* work anymore in v8.

C47D:
> I believe it works in v7 (as the driver is stored at the very beginning of the lv_disp_t structure), however, it does not work anymore in v8.

I had the same impression, I will try it again later today. For v8 I was planning to use the `update_cb` to get the notification of the driver rotation.

C47D:
I might be wrong, do we need to apply https://github.com/lvgl/lvgl/pull/2135/commits/6ac2d9414121b0eb4e96f2eebaa5500773906ae0 to v7?

lv_port/lv_port_display_espressif.c
```diff
@@ -0,0 +1,41 @@
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+
+#include "display_hal.h"
+
+#include "sdkconfig.h"
+#include "driver/gpio.h"
+
+void display_hal_delay(lv_disp_drv_t *drv, uint32_t delay_ms)
```
tore-espressif, line 9: I thought we agreed on `display_port_***` instead of `display_hal_***`
Applies to `display_hal.h` too

lvgl_tft/st7789.h
```diff
@@ -110,11 +99,36 @@ extern "C"
...
 
-void st7789_send_cmd(uint8_t cmd);
-void st7789_send_data(void *data, uint16_t length);
+/**
+ * Control display backlight
+ *
+ * @param drv Pointer to lv_disp_drv_t being used
+ * @param backlight Backlight control
+ */
+void st7789_enable_backlight(lv_disp_drv_t *drv, bool backlight);
```
tore-espressif, line 124: PR https://github.com/lvgl/lvgl_esp32_drivers/pull/95 carved out the backlight control from display driver.
Are you sure want to have backlight and display drivers merged together?

lvgl_tft/display_hal.h
```diff
@@ -0,0 +1,30 @@
...
+/* NOTE We could remove this function from here, because it's used to
+ * initialize the GPIOS on the MCU */
+void display_hal_init_io(void);
+
+/* NOTE drv parameter is meant to be used in multi display projects, so the
+ * user could distinguish multiple displays on their hal implementation */
+void display_hal_delay(lv_disp_drv_t *drv, uint32_t delay_ms);
+void display_hal_backlight(lv_disp_drv_t *drv, uint8_t state);
+void display_hal_gpio_dc(lv_disp_drv_t *drv, uint8_t state);
+void display_hal_gpio_rst(lv_disp_drv_t *drv, uint8_t state);
```
tore-espressif, line 24: I was thinking about single port API that would control GPIO pin.
Like: `display_port_gpio_set(void *some_arg, int pin_num, bool state)`

It could be reused for other pins, not just DC and RST

lvgl_helpers.c
```diff
@@ -105,8 +105,6 @@ void lvgl_driver_init(void)
         DISP_SPI_IO2, DISP_SPI_IO3);
 
     disp_spi_add_device(TFT_SPI_HOST);
-
-    disp_driver_init();
```
tore-espressif, line 109: Is this delete just becouse signature of `disp_driver_init` has changed and the code wasn't updated?

lvgl_tft/disp_driver.c
```diff
@@ -7,7 +7,7 @@
 #include "esp_lcd_backlight.h"
 #include "sdkconfig.h"
 
-void *disp_driver_init(void)
+void *disp_driver_init(lv_disp_drv_t *drv)
```
tore-espressif, line 10: I haven't found a place where this function is called with the new `disp_driver_init(lv_disp_drv_t *drv)` signature

lvgl_tft/st7789.c
```diff
@@ -85,52 +83,68 @@ void st7789_init(void)
...
+    if (rotation != prev_rotation) {
+        /* Update rotation */
+
+        /* TODO Handle all other rotations */
+        if (LV_DISP_ROT_180 == rotation) {
+            st7789_set_orientation(drv, 1);
+        }
+
+        prev_rotation = rotation;
+    }
```
tore-espressif, line 139: I suggest leaving the rotation feature to another PR

lvgl_tft/st7789.c
```diff
@@ -140,73 +154,79 @@ void st7789_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * colo
...
-    };
-
-    ESP_LOGI(TAG, "Display orientation: %s", orientation_str[orientation]);
+#if !defined(ST7789_SOFT_RST)
+    display_hal_gpio_rst(drv, 0);
+    display_hal_delay(drv, 100);
+    display_hal_gpio_rst(drv, 1);
+    display_hal_delay(drv, 100);
+#else
+    st7789_send_cmd(drv, ST7789_SWRESET);
```
tore-espressif, line 224: Pleae add 5ms delay after this call

tore-espressif (commented):
I must admit, that I don't understand all the changes. I have to setup my ST7789 and try it out

tore-espressif:
Sorry for the late reply, I was caught by some other issues.

It seems to me that when we change the `disp_driver_init` function or the lvgl_helpers.c it might affect other drivers too.

BTW, I ordered all LCD types that are currently supported here, plus some more, that were requested from customers. I'd like to setup a testbench, to make sure that all drivers work correctly before we merge some major refactoring into `master`

C47D:
Hi @tore-espressif , don't worry, sorry for all the pings 😸 

Maybe a picture if more descriptive on how I try to achieve this

bloques_st7789

We provide the basic support for ESP targets with the `DISPLAY_ESPRESSIF.c`, if user need to support other configuration or other MCU they remove this file from their build and replace it with their own implementation of `DISPLAY I/O.h` and `DISPLAY IF (INTERFACE).h`

There's a bunch of commits that ended up here because of a rebase from master.

lvgl_tft/st7789.c
C47D: Agree, how should I remove this commit, a revert?

lvgl_tft/disp_driver.c
C47D: In my [test project](https://github.com/lvgl/lv_port_esp32/tree/feat/new_driver_test) I'm calling `st7789_init` and `st7789_flush` directly, planning to remove the disp_driver_init and display_driver_flush all together.

lvgl_helpers.c
C47D: We're initializing the display driver in main, so `lvgl_driver_init` will only initialize the peripherals, we may need to change it's signature.

lvgl_tft/display_hal.h
C47D: It can work for esp tagets, but I'm not sure in other MCU, what do think?

lv_port/lv_port_display_espressif.c
C47D: We can change the name, will do later today.

lvgl_tft/st7789.c
C47D: Fixed with 38bca35

lvgl_tft/st7789.c
C47D: Fixed with fde74d1

lv_port/lv_port_display_espressif.c
C47D: Fixed with d02d309

lvgl_tft/st7789.h
C47D: Fixed with a1a64a9

kisvegabor:
What abut merging it to get a cleaner picture? :slightly_smiling_face:

C47D:
I think I'm waiting for the last feedback. Anyway after merging this I will need to update all the drivers.

lvgl_tft/disp_driver.c
tore-espressif: >  planning to remove the disp_driver_init and display_driver_flush all together.

good to hear!

lvgl_helpers.c
tore-espressif: In that case, all calls to `disp_driver_init` should be deleted from this function, shouldn't they?

lv_port/lv_port_display_espressif.c
```diff
@@ -0,0 +1,41 @@
...
+}
+
+void display_port_backlight(lv_disp_drv_t *drv, uint8_t state)
+{
+    (void) drv;
+
+#ifdef CONFIG_LV_DISP_PIN_BCKL
+    gpio_set_level(CONFIG_LV_DISP_PIN_BCKL, state);
+#endif
+}
```
tore-espressif, line 23: I wanted to use `esp_lcd_backlight.c` for this feature, but we can leave it as it is

tore-espressif:
Also, it might be a good idea to rebase on (or merge from) `develop`, so you get the commit with automated CI build, just to ensure that everything builds correctly

kisvegabor:
I was thinking about what are the future improvements of LVGL that I'd like to focus on and one of them is drivers. 
As a proof of concept I'd like add some significantly different drivers. E.g. `fbdev`, `sdl`, `evdev`, or STM/NXP LCD periphery.
What do you think, how/when/where can I do it?

lv_port/lv_port_display_espressif.c
C47D: We can use the backlight feature.

lvgl_helpers.c
C47D: Yes, they need to be deleted, I will do so later today.

C47D:
> Also, it might be a good idea to rebase on (or merge from) develop, so you get the commit with automated CI build, just to ensure that everything builds correctly

Ok, I will rebase on develop after some commits I have to do. Thanks for the feedback 😺 

> I was thinking about what are the future improvements of LVGL that I'd like to focus on and one of them is drivers.
As a proof of concept I'd like add some significantly different drivers. E.g. fbdev, sdl, evdev, or STM/NXP LCD periphery.
What do you think, how/when/where can I do it?

I think we first need to update all the drivers we have to remove direct dependencies on ESP-IDF (using this abstraction and maybe we need one more), then we could try to add ports for STM/NXP LCD periphery (I would need to get some new hardware for testing). For fbdev, sdl and evdev I only know SDL, the other two I don't, what are they?

lvgl_helpers.c
C47D: Updated in 47756f4

lv_port/lv_port_display_espressif.c
C47D: I've created this initial patch, but I think we should discuss a bit further on how to use `esp_lcd_backlight`.

```diff
diff --git a/lv_port/lv_port_display_espressif.c b/lv_port/lv_port_display_espressif.c
index 9f6f4b1..4ecb218 100644
--- a/lv_port/lv_port_display_espressif.c
+++ b/lv_port/lv_port_display_espressif.c
@@ -13,13 +13,11 @@ void display_port_delay(lv_disp_drv_t *drv, uint32_t delay_ms)
     vTaskDelay(pdMS_TO_TICKS(delay_ms));
 }
 
-void display_port_backlight(lv_disp_drv_t *drv, uint8_t state)
+void display_port_backlight(lv_disp_drv_t *drv, disp_backlight_h *bckl, uint8_t brightness)
 {
     (void) drv;
-
-#ifdef CONFIG_LV_DISP_PIN_BCKL
-    gpio_set_level(CONFIG_LV_DISP_PIN_BCKL, state);
-#endif
+    
+    disp_backlight_set(bckl, brightness);
 }
 
 void display_port_gpio_dc(lv_disp_drv_t *drv, uint8_t state)
diff --git a/lvgl_tft/display_port.h b/lvgl_tft/display_port.h
index 2136ec6..4150e22 100644
--- a/lvgl_tft/display_port.h
+++ b/lvgl_tft/display_port.h
@@ -12,6 +12,8 @@ extern "C"
 #include "lvgl/lvgl.h"
 #endif
 
+#include "esp_lcd_backlight.h"
+
 /**
  * Busy wait delay port
  *
@@ -24,9 +26,10 @@ void display_port_delay(lv_disp_drv_t *drv, uint32_t delay_ms);
  * Backlight control port
  *
  * @param drv Pointer to driver See @ref lv_disp_drv_t
- * @param state State of the backlight signal
+ * @param bckl Pointer to backlight handler See @ref disp_backlight_h
+ * @param brightness Brightness percentage
  */
-void display_port_backlight(lv_disp_drv_t *drv, uint8_t state);
+void display_port_backlight(lv_disp_drv_t *drv, disp_backlight_h *backlight, uint8_t brightness);
 
 /**
  * DC signal control port

```

C47D:
I think I managed to f*ck it up 👎 with the rebase, there's a bunch of duplicated commits.

In another hand, I wasnted to ask @tore-espressif @kisvegabor and more interested people if you think we need a meetup to discuss the new work this repo needs, please let me know.

kisvegabor:
> I think I managed to f*ck it up -1 with the rebase, there's some duplicated commits.

Did you made `git rebase master` on `develop`?

> In another hand, I wasnted to ask @tore-espressif @kisvegabor and more interested people if you think we need a meetup to discuss the new work this repo needs, please let me know.

It's a great idea. There is huge difference in time zones of you, @tore-espressif and me, but I'm sure we can manage it.

C47D:
> Did you made git rebase master on develop?

No, I was in my branch (this one), then `git rebase develop`.

C47D:
> It's a great idea. There is huge difference in time zones of you, @tore-espressif and me, but I'm sure we can manage it.

We can give it a try, you already know my usual schedule for LVGL meetings lol

tore-espressif:
> No, I was in my branch (this one), then git rebase develop. I previously rebased master on my branch, last week I think

At least the CI is passing :D

Hello guys,
I'm definitely in for the planning meeting, however, I'll have to attend to another topics for at least next three weeks. I hope you don't mind postponing this a little.

C47D:
I don't mind, we can maybe merge this and I can work on updating the other drivers while you're busy, what do you think?

kisvegabor (approved):
:+1:

PR #78, opened 2021-06-26
Update KConfig to remove LV_ prefix on orientation checks

jsmestad:
This looks to fix part of issue #50 with the offsets getting overridden.

PR #93, opened 2021-07-30
Fix orientation config define

sidwarkd:
The jd79653a.c file will not compile out of the box because the Kconfig define naming doesn't match the ifdef statements in the source file. This change fixes that.


tore-espressif:
This was merged only 1 month ago: https://github.com/lvgl/lvgl_esp32_drivers/pull/78

We need to agree on the `LV` prefixes before we change it again @C47D

sidwarkd:
@tore-espressif @C47D It appears to me the prefix naming convention was already agreed upon as these orientation ones are the only ones in the Kconfig file that didn't have the LV_ prefix. The reason I found the issue is because the JD79653A driver didn't compile correctly because it looks for LV_ prefixed orientation settings. My guess is that @jsmestad had the opposite issue where whatever driver he is using was looking for non-prefixed defines and these settings were previously prefixed. My suggestion would have been to leave the defines named as before and correct the driver code.

It doesn't really matter to me which prefix convention you would like to use. If this change isn't merged then a different PR should be created to fix the JD7 driver to look for the defines as currently named. However, again, every other config option appears to have the LV_ prefix accept for the orientation ones changed by #78 . So if it's decided to move forward with the LV_ naming convention then we need to search the entire codebase for usages and correct them. Just let me know what you decide and I can either close this PR or move forward with the setting renaming.

tore-espressif:
OK, let's go with the `LV` prefixed version.

The non-prefixed orientation symbols are left only in `ssd1306.c` and `Kconfig` files.

@sidwarkd could you complete this MR?

sidwarkd:
@tore-espressif Yes, I'll update this PR in the next day or two

sidwarkd:
@tore-espressif This PR is ready for review now. Let me know if anything needs to be changed.

tore-espressif:
LGTM!
@C47D do you mind having another look?

PR #96, opened 2021-08-03
Build all display drivers

tore-espressif:
Related: https://github.com/lvgl/lv_port_esp32/pull/287

Build all display drivers when compiling this repository.

Motivation:

- We need to build all drivers to check compile errors in CI
- Building multiple drivers is prerequisite for multi-display control


CMakeLists.txt
```diff
@@ -8,44 +8,24 @@ list(APPEND SOURCES "lvgl_tft/disp_driver.c")
 
 # Include only the source file of the selected
 # display controller.
-if(CONFIG_LV_TFT_DISPLAY_CONTROLLER_ILI9341)
```
C47D, line 11: Thanks for fixing this, it was introduced by me when all the drivers had the same API names, so quite a while ago. This also should make using multiple displays in the same project, isn't?

C47D (approved):
LGTM

CMakeLists.txt
tore-espressif: Yes, compiling all the drivers is prerequisite for multiple displays setup.

But more must be done: currently the (single) display is configured via Kconfig, which is reasonably complex for one display. For multiple displays we would have to dynamically add configuration menus for each enabled display, and then assign the Kconfig options to the correct driver, which is something I wouldn't recommend for number of reasons:
- lower maintainability
- higher complexity
- community knowledge of Kconfig, etc.

That's why I am always for runtime configuration.
(We could still use Kconfig for the naive but frequent usecase of connecting one display with default configuration to the MCU, but we can talk about it later)

C47D: Sure, I agree that having a runtime configuration is a way to go for complex projects.

C47D:
@tore-espressif Do you want to merge this one as well?

tore-espressif:
@C47D  I'm just wondering whether we want to merge this into `master` or we could start new branch like `v2.0.0` for bigger changes, what do you think?

C47D:
> @C47D I'm just wondering whether we want to merge this into master or we could start new branch like v2.0.0 for bigger changes, what do you think?

I think starting a new branch is the way to go, should we also add the `devel` suffix so people know it's under development?

PR #100, opened 2021-08-11
Examples folder and build

tore-espressif:
- Add `examples` folder with first `hello_world` example
- Add LVGL v7.11 submodule: we need it to build the example, plus it gives users clearer message which LVGL version is currently supported
- Add automated build of the example: this gives us immediate feedback whether the repo can be build.  
A lot faster than https://github.com/lvgl/lv_port_esp32/pull/287 where we need to update submodules first

cc @kisvegabor


wreyford:
Yes, I agree with @kisvegabor, why LVGLv7.11 instead of 8. Most of us are benefiting from the LVGL 8 library that brings multiple enhancements, and the drivers  are yet to follow suite.

tore-espressif:
Thank you very much for all the feedback.

> Why v7.11 instead of v8?

v7.11 is the current status of this repository. Of course, we want to move to v8, but we have to make one step at a time.
My reasoning behind this was that it is useful to have automated CI step that checks that the code is compilable _before_ we start changes for LVGL v8.

> but I vote for calling it a test

True. We already have lot of regressions, so I though that this CI test/example_build would help us a lot not to break anything.

> How hard would it be to support both?

Depends what you mean by support :) 
As discussed before I'd like to start with versioning of this repo. Version 1.0.0 would be current `master` with some bugfixes remaining https://github.com/lvgl/lvgl_esp32_drivers/milestone/2 (please add anything that might be missing here), that means LVGL v7.

Version 2.0.0 will aim for LVGL v8 and another breaking changes, so we don't have macros like this
`#if (LVGL_VERSION_MAJOR < 8)`in the code.

Code for v2 should follow standard git workflow e.g. https://gist.github.com/stuartsaunders/448036/5ae4e961f02e441e98528927d071f51bf082662f#git-workflow

kisvegabor:
[Here](https://github.com/lvgl/lv_lib_split_jpg/issues/24) we started talk about releasing and branching difficulties with many repos. In short, it might be reasonable to to use a monorepo. As part of it the drivers could be added directly into the `lvgl` repo. Please, join that conversation and tell what you think.

tore-espressif:
> it might be reasonable to to use a monorepo

Based on the linked discussion, this PR is on the right track to become a monorepo. It is a first step to merge [lv_port_esp32](https://github.com/lvgl/lv_port_esp32) (which is essentially just an example) to this repository.

Also, there is no plan for automated sync between `lvgl_esp32_drivers` and `lvgl`. So if we decide to support new major/minor version of LVGL, there has to be a (rather big) PR updating the drivers, READMEs and examples (all in this repository, with immediate feedback about the build passing or not).

The remaining issues mentioned in the thread are not relevant here (AFAIK)

-  we don't build for multiple platforms, just `esp` 
-  we don't want to track LVGL's `master` branch, just tagged releases that we choose to support

When it comes to merging the drivers to `lvgl`, I know that that is the ultimate goal, but I wanted to have CI, versioning and LVGL v8 support done first.

Please let me know what you think about the priorities and we can change it if need be

kisvegabor:
If the CI testing of for ESP is developed here it will be much easier to add support for other vendors.  

> When it comes to merging the drivers to lvgl, I know that that is the ultimate goal, but I wanted to have CI, versioning and LVGL v8 support done first.

Sounds good to me. Although, the idea of the monorepo just came up, so it's not a plan yet. :slightly_smiling_face: 
To goal was to turn this repo into a multi-platform driver "package" in the end.

tore-espressif:
Are you OK with these changes, can I merge it?

kisvegabor:
The idea of the monorepo seems like a divisive subject so I can't what will be the outcome of the discussion yet. 

Yes, merge it please. This update is important independently of the monorepo.

PR #101, opened 2021-08-15
fix data length overflow

chenghongyao:
change `uint16_t` to `size_t` to prevent length overflow if the area size is too large (like 240x240). I change only one file `st7789.c` but i assume the others should do the same.


C47D:
Thanks for the fix @chenghongyao , I will merge it and try to fix all other display controllers. It's a nice fix, since `send_colors` uses `disp_spi_send_colors(uint8_t *data, size_t length)` and the `length` parameter is `size_t`.

PR #103, opened 2021-08-26
One pixel transformation for ILI948x

### LINKED ISSUE #85: On ESP32C3 with ILI9488 controller at 480x320 16bit, the DMA memory is not sufficient for an entire screen to flush
The function ili9488_flush() is using an algorithm that allocates an entire screen sized DMA buffer, but on the ESP32C3 device, there is not sufficient memory.

I modified the function to do one line at a time, but it might be good to figure out the best of allocation.

I have a modification to this driver that I can share, but it seems that many of the drivers share this flushing code.

Perhaps there is a way to double check the memory required and do the entire screen if available, otherwise drop back to one line or x lines at a time?

tore-espressif:
ILI948x doesn't support 16bit color data in SPI mode, so we have to extrapolate from LVGL's RGB565 (16bit) to 18 bits.

Previously, the whole buffer was transformed, which required a lot of DMA capable RAM. This PR changes to transformation algorithm to process one pixel at a time.

Closes #85

### DIFF
--- a/lvgl_tft/ili9481.c
+++ b/lvgl_tft/ili9481.c
@@ -111,28 +111,6 @@ void ili9481_init(void)
 // Flush function based on mvturnho repo
 void ili9481_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * color_map)
 {
-    uint32_t size = lv_area_get_width(area) * lv_area_get_height(area);
-
-    lv_color16_t *buffer_16bit = (lv_color16_t *) color_map;
-    uint8_t *mybuf;
-    do {
-        mybuf = (uint8_t *) heap_caps_malloc(3 * size * sizeof(uint8_t), MALLOC_CAP_DMA);
-        if (mybuf == NULL)  ESP_LOGW(TAG, "Could not allocate enough DMA memory!");
-    } while (mybuf == NULL);
-
-    uint32_t LD = 0;
-    uint32_t j = 0;
-
-    for (uint32_t i = 0; i < size; i++) {
-        LD = buffer_16bit[i].full;
-        mybuf[j] = (uint8_t) (((LD & 0xF800) >> 8) | ((LD & 0x8000) >> 13));
-        j++;
-        mybuf[j] = (uint8_t) ((LD & 0x07E0) >> 3);
-        j++;
-        mybuf[j] = (uint8_t) (((LD & 0x001F) << 3) | ((LD & 0x0010) >> 2));
-        j++;
-    }
-
     /* Column addresses  */
     uint8_t xb[] = {
         (uint8_t) (area->x1 >> 8) & 0xFF,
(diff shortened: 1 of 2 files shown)


kisvegabor (commented):
Hi,

I found a few things that might be worth to discuss.

C47D:
@tore-espressif Let me know if you want me to test this PR, I got this hardware and an ESP32 dev kit.

SinglWolf:
I tried to apply your patch for my ILI9488 display.
The red color is gone. The picture, which should be in color, appears as monochrome on a blue background (appears on the screen very, very slowly).
On the watch, the colon, which should be blinking in time with the seconds, no longer blinks, but looks like a smooth slideshow from top to bottom. The numbers are updated very slowly too.
The only thing that made me happy was that the [artifacts](https://github.com/lvgl/lvgl_esp32_drivers/issues/116) were gone.

Or have I done something wrong again?

SinglWolf:
After a lot of searching on github, I stumbled upon repo https://github.com/anuprao/esp32_ili9488, which implements something similar for the ILI9488 display controller. Only the version of the LVGL is very old 5.3.0. Maybe this will come in handy for solving the problem?
Fragment of code:
```c
uint8_t buffConv[320*3] = {};

// Used in buffered mode
void ili9488_flush(int32_t x1, int32_t y1, int32_t x2, int32_t y2, const lv_color_t * color_map)
{
	uint8_t data[4];
	
	uint8_t* pbuf;
	uint16_t i, len, chunk;
	//uint8_t r,g,b;
	
	lv_color16_t s;
	lv_color16_t* pColor = color_map;

	/*Column addresses*/
	ili9488_send_cmd(0x2A);
	data[0] = (x1 >> 8) & 0xFF;
	data[1] = x1 & 0xFF;
	data[2] = (x2 >> 8) & 0xFF;
	data[3] = x2 & 0xFF;
	ili9488_send_data(data, 4);

	/*Page addresses*/
	ili9488_send_cmd(0x2B);
	data[0] = (y1 >> 8) & 0xFF;
	data[1] = y1 & 0xFF;
	data[2] = (y2 >> 8) & 0xFF;
	data[3] = y2 & 0xFF;
	ili9488_send_data(data, 4);

	/*Memory write*/
	ili9488_send_cmd(0x2C);
	
	uint32_t size = (x2 - x1 + 1) * (y2 - y1 + 1);

	//Original code
	//ili9488_send_color((void*)color_map, size * 2);	/*Send the remaining data*/
	
	//ili9488_send_color((void*)color_map, size * 2);
	//ili9488_send_color(arrSample, 32);
	
	chunk = 0;
	len = size;
	while(len > 0)
	{
		chunk = (320 < len)? 320: len;
		
		pbuf = buffConv;
		for(i = 0; i<chunk; i++)
		{
			s = *pColor;
			
			*pbuf = s.red << 3;
			pbuf++;
			
			*pbuf = (s.green_h << 3 | s.green_l) << 2;
			pbuf++;
			
			*pbuf = s.blue << 3;
			pbuf++;
			
			pColor++;
		}
		
		ili9488_send_color(buffConv, chunk*3);
		
		len = len - chunk;
	}
	
	printf("size_flush = %d\r\n", size);
	
	//lv_flush_ready();

}
```

tore-espressif:
Hello @SinglWolf ,
Thank you for your patience.
I ordered ILI9488 LCD controller a while ago, so I could test the performance, but the package got stuck at customs.
I will finalize this PR as soon as I get it.
Thanks!

wreyford:
In my experience, this solution helps for the DMA issue, but renders the updates so slow, that the screen is not usable.
I have ordered extra PSRAM to try and solve this issue using hardware - still await that in the mail.

PR #108, opened 2021-09-13
bug fix to use V8 with ILI9341

ogghst:
this to me worked to use the ILI9341 driver on LGVL v8.0. it's rough as i'm a newbie of LVGL but it worked. let me know if you think it's useful.


C47D:
@tore-espressif should we start the v8 support on the drivers?

@ogghst thanks for the contribution, we will see how can we integrate it into the code base.

ogghst:
Thanks for your effort and quick reply. Just fyi I am using this fix on my project that has a complete GUI and works pretty well with almost no other changes compared to the same code on a PC. I will update you in case of issues in this thread.

tore-espressif:
Hello,
it is good to hear that you made it work with so few changes.

Problem with this PR is that updates ILI9341 for LVGL v8, while leaving the rest of the drivers in v7.
I'd prefer having the transition to v8 in one commit/PR.

ogghst:
hi @tore-espressif  absolutely, I agree with you, i'm not in a hurry (for now :) ) and I can wait. Please let me know if I can help you in some way. You guys did an excellent work!

C47D:
It seems to me that most of the suggested changes are now considered in develop branch, thanks for the heads up!

PR #112, opened 2021-09-19
feat(lv_port): Add abstraction for busy signal

C47D:
This signal can be used when driving eink displays. The test for this will be porting the `il3820` display driver to the new driver abstraction layer (with the new capabilities added in this PR).

### DIFF
--- a/lv_port/lv_port_display_espressif.c
+++ b/lv_port/lv_port_display_espressif.c
@@ -40,16 +40,16 @@ void display_port_gpio_rst(lv_disp_drv_t *drv, uint8_t state)
 #endif
 }
 
-display_port_busy_t display_port_gpio_is_busy(lv_disp_drv_t *drv)
+bool display_port_gpio_is_busy(lv_disp_drv_t *drv)
 {
     (void) drv;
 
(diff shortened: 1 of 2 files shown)


C47D:
I worked on this last weekend and I can't get the IL3820 display to work, but it also doesn't work without the changes. So I will try to get it working without the changes first, it's been a while since I used it

C47D:
I will need to buy some stuff to debug the display issue, what about merging this without the display update? that will come later 😄

lv_port/lv_port_display_espressif.c
```diff
@@ -39,3 +39,19 @@ void display_port_gpio_rst(lv_disp_drv_t *drv, uint8_t state)
     gpio_set_level(CONFIG_LV_DISP_PIN_RST, state);
 #endif
 }
+
+display_port_busy_t display_port_gpio_is_busy(lv_disp_drv_t *drv)
```
kisvegabor, line 43: Isn't it enough to return `bool` here?

C47D: It is, do you want me to update it?

C47D: Updated in d627884

kisvegabor (approved):

PR #125, opened 2021-10-05
ili9341: Update driver to use new port interface

C47D:
What the title says.

TODO:

- [ ] Test software reset


kisvegabor (approved):
Looks good to me. It's in harmony with the previous driver update.

Although, I still believe that we should test the architecture/API with unrelated drivers (e.g. ILI9341 and SDL) too to be sure we are on the right track. If it turns out that the API (or the concept) needs to be changed to integrate different kind of drivers all the drivers here needs to be updated again.  

So I suggest updating 2-3 display and indev drivers here and then make experiments with other drivers too.

C47D:
Hi,


This is the driver for ILI9341, did you meant another one?

tore-espressif:
@C47D do you still want me to test the SW reset?

I can check beginning of next week.

C47D:
Hi @tore-espressif,

I think not, it was a connection issue.

kisvegabor:
@C47D 

IMO with IL3820 the display controller based are covered and we can move to an other field (frame buffer, SDL).

> One question, I tried ti run the astyle tool on the source files I've edited. I had to copy over the configuration file into my local repo, but it didn't seem to be working.

What error did you see?

C47D:
> IMO with IL3820 the display controller based are covered and we can move to an other field (frame buffer, SDL).

Where should I open an issue to track that work? On `lv_drivers` ?

kisvegabor:
> Where should I open an issue to track that work? On lv_drivers ?

If we agree that it would beneficial to ship the driver directly with lvgl, I think we can use a branch there. It allows to immediately use the tools of the lvgl repo, including CI, lv_conf_internal.h generation, code formatting, etc. 

You have mentioned that you find it too early so lv_driver can be a consensus. :slightly_smiling_face: 
But if you are ok with the lvgl repo, IMO it'd be best :slightly_smiling_face:

C47D:
> 
> 
> > Where should I open an issue to track that work? On lv_drivers ?
> 
> If we agree that it would beneficial to ship the driver directly with lvgl, I think we can use a branch there. It allows to immediately use the tools of the lvgl repo, including CI, lv_conf_internal.h generation, code formatting, etc.
> 
> You have mentioned that you find it too early so lv_driver can be a consensus. 🙂 But if you are ok with the lvgl repo, IMO it'd be the best 🙂

Well, TBH I'm still "afraid" that incorporating this as-is would break other projects, that's why I've proposed to cleanup this repo first and then see where it belongs.

PR #127, opened 2021-10-05
Fix compilation break when hardware reset is not enabled

### LINKED ISSUE #126: Some drivers doesn't build when no hardware reset is enabled
What the title says, the compilation fails because of those particular drivers use the `CONFIG_LV_DISP_RST` symbol without checking if it exists first.

C47D:
Fixes #126

### DIFF
--- a/lvgl_tft/ili9163c.c
+++ b/lvgl_tft/ili9163c.c
@@ -69,7 +69,7 @@
 
 #define ST77XX_MADCTL_MY 0x80
 #define ST77XX_MADCTL_MX 0x40
-#define ST77XX_MADCTL_MV 0x20 #define
+#define ST77XX_MADCTL_MV 0x20
 #define ST77XX_MADCTL_ML 0x10
 #define ST77XX_MADCTL_RGB 0x00
 #define ST77XX_MADCTL_BGR 0x08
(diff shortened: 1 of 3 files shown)


lvgl_tft/jd79653a.c
```diff
@@ -481,3 +473,14 @@ void jd79653a_init()
 
     ESP_LOGI(TAG, "Panel is up!");
 }
+
+static void jd79653a_reset(void)
+{
+#if defined CONFIG_LV_DISP_PIN_RST
+    gpio_set_level(PIN_RST, 0);
```
tore-espressif, line 480: Is this pin configured as output, before setting its level?

lvgl_tft/uc8151d.c
```diff
@@ -266,3 +259,14 @@ void uc8151d_init()
     uc8151d_panel_init();
     ESP_LOGI(TAG, "Panel initialised");
 }
+
+static void uc8151d_reset(void)
+{
+#if defined CONFIG_LV_DISP_USE_RST
+    gpio_set_level(PIN_RST, 0);
```
tore-espressif, line 266: Is this pin configured as output, before setting its level?

lvgl_tft/ili9163c.c
```diff
@@ -246,3 +236,14 @@ static void ili9163c_set_orientation(uint8_t orientation)
 	ili9163c_send_cmd(ILI9163C_MADCTL);
 	ili9163c_send_data((void *)&data[orientation], 1);
 }
+
+static void ili9163c_reset(void)
+{
+#if CONFIG_LV_DISP_USE_RST
+    gpio_set_level(ILI9163C_RST, 0);
```
tore-espressif, line 243: Is this pin configured as output, before setting its level?

tore-espressif (commented):
I don't see where the reset pin is configured. Otherwise LGTM!

PR #129, opened 2021-10-07
Update (initialization) helpers

### LINKED ISSUE #128: Helper for GPIO configuration
The new drivers init routines are getting rid of the GPIO configuration, we should add a user helper to configure the GPIOs used by the driver, similar to the peripherals (SPI/I2C) helper API.

C47D:
Adds helper function to initialize GPIOs, this avoids having to configure and initialize GPIOs in the drivers init function.

Closes #128

Any suggestions with the name of the function are welcome. Same with `lvgl_driver_init`, now it only initializes the SPI/I2C peripheral.

### DIFF
--- a/lvgl_helpers.c
+++ b/lvgl_helpers.c
@@ -134,6 +134,39 @@ void lvgl_driver_init(void)
 #endif
 }
 
+void display_bsp_init_io(void)
+{
+    esp_err_t err = ESP_OK;
+    gpio_config_t io_conf;
+
+#ifdef CONFIG_LV_DISPLAY_USE_DC
+    io_conf.mode = GPIO_MODE_OUTPUT;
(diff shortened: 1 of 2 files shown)


kisvegabor:
Is it related to the new architecture?

C47D:
Kind of, I'm removing the GPIO initialization from the display drivers to a more generic way.

So at the end there would not be any GPIO specific code in the display driver init function.

kisvegabor:
I think it's a great direction. I was also thinking about it, and concluded that GPIOs can be handled in the display_send, indev_read functions too along with the  SPI/I2C peripheries.

kisvegabor:
DC pin also can be managed in the SPI driver.

kisvegabor:
Isn't it? :smile: 

E.g. a `send_cmd` can handle the DC pin too. Or have I missed something?

PR #142, opened 2021-11-14
PCD8544 driver

mringwal:
Driver for the PCD8544 Controller - commonly referred as Nokia 3110/5110 display.

I just started with LVGL and based the code on other drivers as well as the Adafruit and u8g2 libraries and the example shows "Hello world" on the LCD as expected.

The driver assumes a resolution of 84x48 (physical landscape). I didn't understand how rotation should be handled. For monochrome drivers, it seems like pcd8544_set_px_cb is called for every pixel to set. In this case, a simple way to implement rotation would be to just rotate the coordinates in the set pixel function.


lvgl_tft/pcd8544.h
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+#ifndef PCD8544_H
+#define PCD8544_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    /*********************
```
C47D, line 14: Could you remove the tabs before this line? Same for lines below

lvgl_tft/pcd8544.c
```diff
@@ -0,0 +1,146 @@
...
+
+void pcd8544_flush(lv_disp_drv_t * disp_drv, const lv_area_t * area, lv_color_t * color_map){
+
+    pcd8544_send_cmd(0x20);     /* activate chip (PD=0), horizontal increment (V=0), enter extended command set (H=0) */
+
+    uint8_t * buf = (uint8_t *) color_map;
+
+    if ((area->x1 == 0) && (area->y1 == 0) && (area->x2 == (disp_drv->hor_res - 1)) && (area->y2 == (disp_drv->ver_res - 1))){
+
+        // optimize flush of complete frame buffer in a single SPI transaction
```
C47D, line 121: Is this a TODO comment? If so, could you add TODO at the beginning?

C47D:
Hi @mringwal 

It does look nice overall, thanks for the submission. I left a couple of comments/questions, I can test this this weekend.

lvgl_tft/pcd8544.h
mringwal: Good point. No idea why they are indented.

lvgl_tft/pcd8544.c
mringwal: It is an explanation. I'll improve it.

tore-espressif (commented):
LGTM! Code looks OK, but I don't have this display controller, so I couldn't try it

C47D:
I think I have it, I will try to find it and let you know how it works.

C47D:
I was able to build the project but the display didn't showed the expected message. I will check my wiring as I'm using some jumpers I found laying around.

Just to confirm the expected configuration, please let me know if something is wrong.

- 1bit color depth
- primary color: 0x000000
- secondary color: 0xffffff
- horizontal size: 84
- vertical size: 48
- mono theme

mringwal:
Hi @C47D - thanks for trying. I just had a jumper wire yesterday that was just broken and couldn't get a simple buzzer to work (only 2 wires, just send some square wave and it should work...)

Configuration: i've started with the main demo and configured the SPI, resolution, 1 byte color depth, and the mono theme.
Your screen size and mono theme are correct.

Hm... I have "1 byte per pixel"  as LVGL Configuration -> Color Depth. I wondered about that as the my code and the existing other drivers did use a 1-bit bitmap).

I have 0xff00000 theme default primary color and 0x0000ff as secondary - as it was the default.

Finally, the display has a 'contrast' value. If it's too high, you'll get a black screen where you can read the text if you're looking very carefully.  It's this line:
 pcd8544_send_cmd(0xc0);     /* medium Vop = Contrast 0x40 = 64 */

(There's also a pcd8544_set_contrast(..) function that changes it. maybe you can try a sweep over the range)

Thanks!

C47D:
> Hi @C47D - thanks for trying. I just had a jumper wire yesterday that was just broken and couldn't get a simple buzzer to work (only 2 wires, just send some square wave and it should work...)
> 
> Configuration: i've started with the main demo and configured the SPI, resolution, 1 byte color depth, and the mono theme. Your screen size and mono theme are correct.
> 
> Hm... I have "1 byte per pixel" as LVGL Configuration -> Color Depth. I wondered about that as the my code and the existing other drivers did use a 1-bit bitmap).
> 
> I have 0xff00000 theme default primary color and 0x0000ff as secondary - as it was the default.
> 
> Finally, the display has a 'contrast' value. If it's too high, you'll get a black screen where you can read the text if you're looking very carefully. It's this line: pcd8544_send_cmd(0xc0); /* medium Vop = Contrast 0x40 = 64 */
> 
> (There's also a pcd8544_set_contrast(..) function that changes it. maybe you can try a sweep over the range)
> 
> Thanks!

You're right, it's 1byte color depth, not as I mentioned. I will play with the contrast value, all I see is a black screen 😄

C47D:
I'm not being able to replicate it, can you attach the generated `sdkconfig.h` file? It's in the `build/config` directory of the project.

mringwal:
Sure. I've attached the `sdkconfig.`h (from `build/include/config`).

I'm using the master version of lv_port_esp32 with the example and also had to added the CONFIG_LV_TFT_DISPLAY_CONTROLLER_PCD8544 in this block.

```
diff` --git a/main/main.c b/main/main.c
index 388008a..2968f06 100644
--- a/main/main.c
+++ b/main/main.c
@@ -100,7 +100,8 @@ static void guiTask(void *pvParameter) {
 #if defined CONFIG_LV_TFT_DISPLAY_CONTROLLER_IL3820         \
     || defined CONFIG_LV_TFT_DISPLAY_CONTROLLER_JD79653A    \
     || defined CONFIG_LV_TFT_DISPLAY_CONTROLLER_UC8151D     \
-    || defined CONFIG_LV_TFT_DISPLAY_CONTROLLER_SSD1306
+    || defined CONFIG_LV_TFT_DISPLAY_CONTROLLER_SSD1306     \
+    || defined CONFIG_LV_TFT_DISPLAY_CONTROLLER_PCD8544

     /* Actual size in pixels, not bytes. */
     size_in_px *= 8;
```
[sdkconfig.h.zip](https://github.com/lvgl/lvgl_esp32_drivers/files/7600999/sdkconfig.h.zip)

C47D:
Thanks, I will update the main.c file. I guess we could update that code section, so we don't have to edit it after adding support for new display drivers.

mringwal:
Did you get the display to work on your side?

C47D:
> Did you get the display to work on your side?
> And yes, a way to add drivers without modifying main.c would be nice. I think the list of devices matches the ones that are monochrome?

Yes, got it working.
! [IMG_20211125_171230.jpg](https://user-images.githubusercontent.com/11037705/143506846-b064d235-0272-4d6b-983b-4752fe401af7.jpg)

C47D:
> Did you get the display to work on your side?
> And yes, a way to add drivers without modifying main.c would be nice. I think the list of devices matches the ones that are monochrome?

Yes, got it working with the patch!

mringwal:
Thanks for testing/validating!

Issue #143, opened 2021-11-19
Communication interface abstraction


C47D:
Hi,

As we've discussed before there's the desire to abstract the communication interface (SPI, I2C, Parallel, etc.) in the display drivers.

## Current implementation

We only have support for SPI and I2C. Here's how currently (develop branch) a command and data are being sent over SPI (ili9341):
```c
static inline void set_cmd_mode(lv_disp_drv_t * drv)
{
    display_port_gpio_dc(drv, 0);
}

static inline void set_data_mode(lv_disp_drv_t * drv)
{
    display_port_gpio_dc(drv, 1);
}

static void ili9341_send_cmd(lv_disp_drv_t * drv, uint8_t cmd)
{
    disp_wait_for_pending_transactions();
    set_cmd_mode(drv);
    disp_spi_send_data(&cmd, 1);
}

static void ili9341_send_data(lv_disp_drv_t *drv, void * data, uint16_t length)
{
    disp_wait_for_pending_transactions();
    set_data_mode(drv);
    disp_spi_send_data(data, length);
}

static void ili9341_send_color(lv_disp_drv_t *drv, void * data, uint16_t length)
{
    disp_wait_for_pending_transactions();
    set_data_mode(drv);
    disp_spi_send_colors(data, length);
}
```

Here's how they're sent over I2C:
```c
static uint8_t send_data(lv_disp_drv_t *disp_drv, void *bytes, size_t bytes_len)
{
    (void) disp_drv;

    uint8_t *data = (uint8_t *) bytes;

    return lvgl_i2c_write(OLED_I2C_PORT, OLED_I2C_ADDRESS, data[0], data + 1, bytes_len - 1 );
}

static uint8_t send_pixels(lv_disp_drv_t *disp_drv, void *color_buffer, size_t buffer_len)
{
    (void) disp_drv;

    return lvgl_i2c_write(OLED_I2C_PORT, OLED_I2C_ADDRESS, OLED_CONTROL_BYTE_DATA_STREAM, color_buffer, buffer_len);
}
```

In the SPI case we have two functions to send data: `disp_spi_send_data` and `disp_spi_send_colors`, defined in `disp_spi.h`:
```c
static inline void disp_spi_send_data(uint8_t *data, size_t length) {
    disp_spi_transaction(data, length, DISP_SPI_SEND_POLLING, NULL, 0, 0);
}

static inline void disp_spi_send_colors(uint8_t *data, size_t length) {
[...]

PR #144, opened 2021-11-20
Communication interface proposal

C47D:
See #143 for additional context

TODO:

- [x] Add abstraction to I2C driven display.
- [ ] Add error returns

### DIFF
--- a/lv_port/lv_port_display_espressif.c
+++ b/lv_port/lv_port_display_espressif.c
@@ -66,48 +66,53 @@ bool display_port_gpio_is_busy(lv_disp_drv_t *drv)
     return device_busy;
 }
 
-void display_interface_send_cmd(lv_disp_drv_t *drv, uint8_t cmd, void *args, size_t args_len)
+void display_interface_send_cmd(lv_disp_drv_t *drv, uint32_t cmd, cmd_width_t cmd_width, void *args, size_t args_len)
 {
     (void) drv;
 
 #if defined (CONFIG_LV_TFT_DISPLAY_PROTOCOL_SPI)
     disp_wait_for_pending_transactions();
     display_port_gpio_dc(drv, LV_DISPLAY_DC_CMD_MODE);
-    disp_spi_send_data(&cmd, 1);
+
+    if (CMD_WIDTH_8BITS == cmd_width) {
+        disp_spi_send_data(&cmd, 1);
+    }
+    else if (CMD_WIDTH_16BITS == cmd_width) {
+        /* Send 16bits cmd */
+    }
+    else {
+        /* Unsupported cmd size */
+    }
 
     if (args != NULL) {
         display_port_gpio_dc(drv, LV_DISPLAY_DC_DATA_MODE);
         disp_spi_send_data(args, args_len);
     }
 #elif defined (CONFIG_LV_TFT_DISPLAY_PROTOCOL_I2C)
-    (void) drv;
     uint8_t *data = (uint8_t *) args;
+    
     lvgl_i2c_write(OLED_I2C_PORT, OLED_I2C_ADDRESS, cmd, data, args_len); 
 #endif
 }
 
-void display_interface_send_data_async(lv_disp_drv_t *drv, void *data, size_t len)
+void display_interface_send_data(lv_disp_drv_t *drv, void *data, size_t len, data_xfer_mode_t mode)
 {
-#if defined (CONFIG_LV_TFT_DISPLAY_PROTOCOL_SPI)
-    disp_wait_for_pending_transactions();
-    display_port_gpio_dc(drv, LV_DISPLAY_DC_DATA_MODE);
-    
-    disp_spi_send_colors(data, len);
-#elif defined (CONFIG_LV_TFT_DISPLAY_PROTOCOL_I2C)
     (void) drv;
-    lvgl_i2c_write(OLED_I2C_PORT, OLED_I2C_ADDRESS, OLED_CONTROL_BYTE_DATA_STREAM, data, len);
-#endif
-}
 
-void display_interface_send_data(lv_disp_drv_t *drv, void *data, size_t len)
-{
 #if defined (CONFIG_LV_TFT_DISPLAY_PROTOCOL_SPI)
     disp_wait_for_pending_transactions();
     display_port_gpio_dc(drv, LV_DISPLAY_DC_DATA_MODE);
     
-    disp_spi_send_colors(data, len);
+    if (DATA_XFER_MODE_ASYNC == mode) {
+        disp_spi_send_colors(data, len);
+    }
+    else if (DATA_XFER_MODE_SYNC == mode) {
+        /* Send data in polling mode and call lv_disp_flush after that */
+    }
+    else {
+
+    }
 #elif defined (CONFIG_LV_TFT_DISPLAY_PROTOCOL_I2C)
-    (void) drv;
     lvgl_i2c_write(OLED_I2C_PORT, OLED_I2C_ADDRESS, OLED_CONTROL_BYTE_DATA_STREAM, data, len);
 #endif
 }
--- a/lv_port/lv_port_display_espressif.c
+++ b/lv_port/lv_port_display_espressif.c
@@ -6,6 +6,11 @@
 #include "sdkconfig.h"
 #include "driver/gpio.h"
 
+#include "disp_spi.h"
+
+#define LV_DISPLAY_DC_CMD_MODE    0
+#define LV_DISPLAY_DC_DATA_MODE   1
+
 void display_port_delay(lv_disp_drv_t *drv, uint32_t delay_ms)
 {
     (void) drv;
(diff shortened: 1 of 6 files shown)


C47D:
Hi @ropg,

I wanted to ask your thoughts on the I2C part of the abstraction. Is there something we could do better?

ropg:
Hey there...  :)

I can't really figure out what's going on and why from reading #143 and the code in here. Most pressing question: where is:

```c
/* TODO: This is ssd1306 specific */
#define OLED_I2C_PORT                       (CONFIG_LV_I2C_DISPLAY_PORT)
#define OLED_I2C_ADDRESS                    0x3C
```

supposed to end up?

Abstracting SPI and I2C to use a common interface is hard because of all the differences: ESP32 does locking natively, there's many more modes that can be used concurrently, etc. It really depends on a lot of factors. I'm a bit pressed for time, but I think re-reading the thread that led to the introduction of I2C Manager makes sense because many things were discussed at length there. As for changes to how I2C is done: if I'm clearer on what the goal is, I can better answer the question as to whether I see a way to do things better.

C47D:
Hi,

Thanks for the quick answer, I should have added more context.

The final goal is to have drivers with no MCU specific code directly into the main LVGL repository. 

The proposal is to have a `display_port` header file that defines a set of generic functions used inside the drivers. 

The implementation of those functions, e.g. the ESP32 specific implementation, is kept in the `lv_port` directory in only one source file.
More MCU example implementations can be added in their own source files.

Users can define their customized implementation on their source files.

This particular PR moves the existing SPI and I2C code behind that generic API.

ropg:
I see. I guess that is nice in many ways but also opens some specific cans of worms as discussed in the I2C Manager PR thread. I like the part of the more generic drivers, just something needs to define how it's all hooked together. Maybe the bits that stay in the specific hardware repositories also need to define some kind of interface for that side. Maybe a generic pointer that you pass to the display driver that it doesn't need to understand, but that points to a struct that holds whatever it is that driver needs to talk to the world, i.e. the data needed by the implementations of the generic function calls. So in the case of that display, it would point to a struct that merely holds the I2C port number, 0 or 1 because the Kconfig for I2C Manager takes care of pins, speed, etc. In other cases (SPI?), it might hold the SPI mode, speed, possibly pins, etc ?

But it really also depends on what the architecture for the common part looks like.

lv_port/lv_port_display_espressif.c
```diff
@@ -66,48 +66,53 @@ bool display_port_gpio_is_busy(lv_disp_drv_t *drv)
     return device_busy;
 }
 
-void display_interface_send_cmd(lv_disp_drv_t *drv, uint8_t cmd, void *args, size_t args_len)
+void display_interface_send_cmd(lv_disp_drv_t *drv, uint32_t cmd, cmd_width_t cmd_width, void *args, size_t args_len)
```
kisvegabor, line 69: Just my 2 cents: I'd use a simple `uint8_t` instead of `cmd_width_t`.

lvgl_tft/display_port.h
```diff
@@ -15,6 +15,17 @@ extern "C"
...
+typedef enum {
+    CMD_WIDTH_8BITS,
+    CMD_WIDTH_16BITS,
+    CMD_WIDTH_INVALID,
+} cmd_width_t;
+
+typedef enum {
+    DATA_XFER_MODE_SYNC,
+    DATA_XFER_MODE_ASYNC,
+} data_xfer_mode_t;
```
kisvegabor, line 27: Why do we need to differentiate sync and async transfer at this level? Why can't the platform-specific driver decide how to send?

C47D: Removed

lv_port/lv_port_display_espressif.c
C47D: Done, handled as lvgl does

lvgl_tft/display_port.h
kisvegabor: Thanks! 

One more thing came to my mind: the prototype should be something like:
```c
void display_interface_update_image(lv_disp_drv_t *drv, lv_arae_t * area, lv_color_t * data);
```

It's because in case of devices with pure frame buffer(s), without a display controller, we can't send commands to set the "window". Instead they want to swap the frame buffers, memcpy/DMA the rendered image, etc.

So this command should tell the area too.

C47D: I think I understand you now, what if we add another display port API for those kind of displays instead of trying to have generic one?

Let me try to confirm my understanding with an example.

We have N display types:
- An ILI9341
- Pure frame buffer

With the ILI9341 we need to let the display know the area coordinates with commands and then send the color data, like so:
```c
void ili9341_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * color_map)
{
    uint8_t data[4] = {0};
    uint32_t size = lv_area_get_width(area) * lv_area_get_height(area);

    /*Column addresses*/
    data[0] = (area->x1 >> 8) & 0xFF;
    data[1] = area->x1 & 0xFF;
    data[2] = (area->x2 >> 8) & 0xFF;
    data[3] = area->x2 & 0xFF;
    
    display_interface_send_cmd(drv, 0x2A, CMD_WIDTH_8BITS, data, sizeof(data));

    /* Page addresses */
    data[0] = (area->y1 >> 8) & 0xFF;
    data[1] = area->y1 & 0xFF;
    data[2] = (area->y2 >> 8) & 0xFF;
    data[3] = area->y2 & 0xFF;
    
    display_interface_send_cmd(drv, 0x2B, CMD_WIDTH_8BITS, data, sizeof(data));

    /* Memory write */
    display_interface_send_cmd(drv, 0x2C, CMD_WIDTH_8BITS, NULL, 0);
    display_interface_send_data(drv, color_map, size * 2);
}
```

With the pure frame buffer display:
```c
void pure_frame_buffer_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * color_map)
{
    display_interface_send_frame_buffer(drv, area, color_map); // equivalent to your suggestion
}
```

---

If we make a generic function, the flush callbacks would end up being just wrappers of it.
```c
void ili9341_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * color_map)
{
    display_interface_update_image(drv, area, color_map);
}
```

```c
void pure_frame_buffer_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * color_map)
{
    display_interface_update_image(drv, area, color_map);
}
```

What do you think?

kisvegabor: I vote for generic, because... it's more generic :laughing: I think it's better to have one function instead of two similar. The users can ignore the `area` parameter if they don't use it.

C47D: I'm not totally convinced to be honest, I think the `display_interface_update_image` function would end up having a lot of code inside. Can we wait for more opinions? What do you think?

kisvegabor: >I think the display_interface_update_image function would end up having a lot of code inside.  

Why will it have more code than `display_interface_send_data`?

C47D: So, this would be the flush callback of the ssd1306 display driver:

```c
void ssd1306_flush(lv_disp_drv_t * disp_drv, const lv_area_t * area, lv_color_t * color_p)
{
    /* Divide by 8 */
    uint8_t row1 = area->y1 >> 3;
    uint8_t row2 = area->y2 >> 3;

    uint8_t conf[] = {
        OLED_CONTROL_BYTE_CMD_STREAM,
        OLED_CMD_SET_MEMORY_ADDR_MODE,
        0x00,
        OLED_CMD_SET_COLUMN_RANGE,
        (uint8_t) area->x1,
        (uint8_t) area->x2,
        OLED_CMD_SET_PAGE_RANGE,
        row1,
        row2,
    };

    display_interface_send_cmd(disp_drv, OLED_CONTROL_BYTE_CMD_STREAM, &conf[1], sizeof(conf) - 1);
    display_interface_send_data(disp_drv, color_p, OLED_COLUMNS * (1 + row2 - row1));
    lv_disp_flush_ready(disp_drv);
}
```

If we use the `display_interface_send_image` interface we would need to handle the specifics of ssd1306 inside, rather than in the flush callback:

```c
void ssd1306_flush(lv_disp_drv_t * disp_drv, const lv_area_t * area, lv_color_t * color_p)
{
    display_interface_send_image(drv, area, color_p);
}

void display_interface_send_image(lv_disp_drv_t *drv, lv_area_t *area, lv_color_t *color)
{
    if (display_is_ssd1306) {
    /* Divide by 8 */
    uint8_t row1 = area->y1 >> 3;
    uint8_t row2 = area->y2 >> 3;

    uint8_t conf[] = {
        OLED_CONTROL_BYTE_CMD_STREAM,
        OLED_CMD_SET_MEMORY_ADDR_MODE,
        0x00,
        OLED_CMD_SET_COLUMN_RANGE,
        (uint8_t) area->x1,
        (uint8_t) area->x2,
        OLED_CMD_SET_PAGE_RANGE,
        row1,
        row2,
    };

    display_interface_send_cmd(disp_drv, OLED_CONTROL_BYTE_CMD_STREAM, &conf[1], sizeof(conf) - 1);
    display_interface_send_data(disp_drv, color_p, OLED_COLUMNS * (1 + row2 - row1));

    lv_disp_flush_ready(disp_drv);
    }
}
```

kisvegabor: I meant the end of the `ssd1306_flush` looks like this:

```c
    display_interface_send_cmd(disp_drv, OLED_CONTROL_BYTE_CMD_STREAM, &conf[1], sizeof(conf) - 1);
    display_interface_send_image(disp_drv, area, color_p);
```

So you give the "window commands" and then ask to send the image too. Finally, only `display_interface_send_data` needs to be called. With "pure frame buffer" display `display_interface_send_cmd` won't be called only `display_interface_send_image`.

C47D: Do you mean to implement something like this?

```c
void display_interface_send_image(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * color_map)
{
    if (is_pure_frame_buffer(drv)) {
        pure_frame_buffer_flush(drv, area, color_map);
    }
    else {
        display_interface_send_data(drv, color_map, /* Calculate bytes to send from area */);
    }
}
```

If not, would you care to have a sketch of how `display_interface_send_image` would look inside?

kisvegabor: Yes, exactly. To complete your example:
```c
uint32_t s = lv_area_get_size(area) * sizeof(lv_color_t);
display_interface_send_data(drv, color_map, s);
```

What do you think?

(I'd call it `display_interface_update_image` as it's not sure it send anything, but it's a question detail)

C47D: Is there a configuration value we can look for to know if a driver is a pure frame buffer display?

kisvegabor: Not really, only its flush_cb does its job differently.

C47D:
Closed in favor of #180

PR #154, opened 2021-12-22
add offset for 240x135 displays based on ST7789

arktrin:


arktrin:
52px and 53px offsets are not an error. Tested on my 240x135 displays that came from different places.
I'm also not sure if this part of Kconfig is right for "TTGO T-Display" board:
```
config LV_TFT_DISPLAY_X_OFFSET
    depends on LV_TFT_DISPLAY_OFFSETS
    int "X offset"
    default 40 if LV_PREDEFINED_DISPLAY_TTGO && (LV_DISPLAY_ORIENTATION_LANDSCAPE || LV_DISPLAY_ORIENTATION_LANDSCAPE_INVERTED)
    default 53 if LV_PREDEFINED_DISPLAY_TTGO && (LV_DISPLAY_ORIENTATION_PORTRAIT  || LV_DISPLAY_ORIENTATION_PORTRAIT_INVERTED)
    default 0

config LV_TFT_DISPLAY_Y_OFFSET
    depends on LV_TFT_DISPLAY_OFFSETS
    int "Y offset"
    default 53  if LV_PREDEFINED_DISPLAY_TTGO && (LV_DISPLAY_ORIENTATION_LANDSCAPE || LV_DISPLAY_ORIENTATION_LANDSCAPE_INVERTED)
    default 40 if LV_PREDEFINED_DISPLAY_TTGO && (LV_DISPLAY_ORIENTATION_PORTRAIT  || LV_DISPLAY_ORIENTATION_PORTRAIT_INVERTED)
    default 0
```
Can't confirm because I don't have this board but I'm sure it won't work for my displays.

lvgl_tft/st7789.c
```diff
@@ -139,13 +139,29 @@ void st7789_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * colo
...
+        (CONFIG_LV_DISPLAY_ORIENTATION_PORTRAIT_INVERTED)
+        offsetx1 += 40;
+        offsetx2 += 40;
+        offsety1 += 53;
+        offsety2 += 53;
+    #endif
+#elif (LV_HOR_RES_MAX == 135) && (LV_VER_RES_MAX == 240)
+    #if (CONFIG_LV_DISPLAY_ORIENTATION_LANDSCAPE) || \
+        (CONFIG_LV_DISPLAY_ORIENTATION_LANDSCAPE_INVERTED)
+        offsetx1 += 52;
```
C47D, line 160: would this be 53?

lvgl_tft/st7789.c
```diff
@@ -139,13 +139,29 @@ void st7789_flush(lv_disp_drv_t * drv, const lv_area_t * area, lv_color_t * colo
...
+        offsetx2 += 52;
```
C47D, line 161: same here

C47D:
Hi,

Thanks for the updates, I only have the 240*240 display so that's the only one I considered.

> I'm also not sure if this part of Kconfig is right for "TTGO T-Display" board

I'm not sure either but as far as I remember it was a contribution, so we assumed it's working as expected (I don't have that kit)

@tore-espressif Should we start thinking about a way to avoid using the `LV_HOR` and `LV_VER` symbols as they were removed in v8?

lvgl_tft/st7789.c
arktrin: This is not a mistake. I don't know why it is working like that. Tested in all display orientations on my 240x135 displays that came from different places.

lvgl_tft/st7789.c
arktrin: This is not a mistake. I don't know why it is working like that. Tested in all display orientations on my 240x135 displays that came from different places.

arktrin:
> Having all those ifdef out of the flush callback would be nice. I'm thinking how can we do that.

I was kind of struggling not to PR version where all those ifdefs are removed from st7789.c and let user set those offsets by themselves in menuconfig, maybe with a bit of a hint or something. Not being able to get those LV_HOR and LV_VER from lvgl itself stopped me. Should I PR this version to let you decide which one is better?

C47D:
Hi,

No, let's merge this as is and figure out how to clean up later. Only add a comment about the 52 and 53 values being used somewhere, maybe at the top of the flush callback. I don't recall seeing it.

tore-espressif:
Hi guys,

yep, the offsets can be different on displays using the same controller, even the ones that have same vendor and resolution (!).
So we need to make this configurable, but for the time being we can hard-code it for some commonly used displays. 

> Should we start thinking about a way to avoid using the LV_HOR and LV_VER symbols

I had a branch that was trying to do this, but these macros are almost everywhere in our drivers :'(

C47D:
@tore-espressif yes, I have an idea on how to start removing them, do we have an issue for v8 compatibility? Maybe we can add a checklist to it and list this 'issues'

arktrin:
> Only add a comment about the 52 and 53 values being used somewhere, maybe at the top of the flush callback

Done!

C47D (approved):
LGTM!, thanks again for the contribution @arktrin

PR #156, opened 2021-12-24
Update lvgl helpers

### LINKED ISSUE #155: [develop] SPI_HOST_MAX not defined
Using ESP-IDF version 4.4, compilation fails because of the missing `SPI_HOST_MAX` symbol.

C47D:
Updates both lvgl_helpers files, if you have something else in mind please let me know.

Closes #155


C47D:
@tore-espressif What version of LVGL is the GA pulling? I'm having no errors when building the lv_port_esp32 (LVGL v7) with this branch.

embeddedt:
It's pulling the version [here](https://github.com/lvgl/lvgl_esp32_drivers/tree/develop/examples/common_components) which is v8...

C47D:
Thanks, time to think how to handle v8 then 😄 

 `DISP_BUF_SIZE` is being used to calculate the draw buffer size, for v7 we use the `LV_HOR_RES_MAX` and `LV_VER_RES_MAX` LVGL configuration values. For v8 both of those configuration values were removed, so we would need to get those values from somewhere else. @tore-espressif do you think we could get them using the Kconfig interface or from somewhere else (defined by user in their program)?

**EDIT**
@embeddedt how are the `physical_hor_res` and `physical_ver_res` members of `lv_disp_drv_t` being used? From my understanding they're used when not all of the display is used by LVGL(?)

@tore-espressif is it ok if we update the example to v8? Seems like it was written with v7 in mind, or maybe have a build for LVGL v7 and another for LVGL v8.

C47D:
The compilation is now only failing because of the missing migration from v7 to v8.

But in order to calculate the display buffer size we would need to get the display size from somewhere else, it might be Kconfig menu (display configuration not LVGL)

tore-espressif:
@C47D Sorry, I was on vacation, I'll get into it next week

C47D:
@tore-espressif don't worry, I'll try to fix the errors on the drivers and also squash some commits, I'm currently using the web editor.

C47D:
So I needed more changes than expected, specially in the drivers that depends on symbols such as `LV_HOR_RES_MAX`, `LV_VER_RES_MAX`, `DISP_BUF_SIZE`, etc., but I think we're going in the right direction. Having those CI builds helped me a lot, because my personal PC takes a lot of time per build, so it's faster for me to upload the commit and see the CI build failing.

On Sunday I will squash some commits and get this ready for review.

PR #162, opened 2022-01-04
Display Kconfig: Fix TTGO default SPI bus and display orientation checks

C47D:
Fix based on https://github.com/lvgl/lv_port_esp32/issues/289#issuecomment-1001704529


lvgl_tft/Kconfig
```diff
@@ -469,14 +469,15 @@ menu "LVGL TFT Display controller"
 
     choice
         prompt "TFT SPI Bus." if LV_TFT_DISPLAY_PROTOCOL_SPI
-        default LV_TFT_DISPLAY_SPI2_HOST
-        help
-              Select the SPI Bus the TFT Display is attached to.
+        default LV_TFT_DISPLAY_SPI2_HOST if LV_PREDEFINED_DISPLAY_TTGO && !IDF_TARGET_ESP32S2
+        default LV_TFT_DISPLAY_SPI3_HOST if !LV_PREDEFINED_DISPLAY_TTGO && IDF_TARGET_ESP32S2
```
arktrin: Why do you choose exactly SPI3_HOST for all ESP32S2 targets?

C47D: This PR is based on a patch received on the linked comment. Do you think it would be better to choose another SPI? I'm not familiar with that particular device

arktrin: As of datasheet and online docs esp32s2 do have two working general purpose SPI: SPI2 and SPI3.  I'm just wondering why SPI3_HOST is the default choice for esp32s2. Also TTGO T Display board have esp32 (not esp32s2), that's why `!LV_PREDEFINED_DISPLAY_TTGO && IDF_TARGET_ESP32S2` should be always true if esp32s2 is target,  unless someone is purposely picking esp32s2 as target and want to use `LV_PREDEFINED_DISPLAY_TTGO`, but pin configuration is hardcoded and cannot be changed  if `LV_PREDEFINED_DISPLAY_TTGO` is picked.

lvgl_tft/Kconfig
```diff
@@ -469,14 +469,15 @@ menu "LVGL TFT Display controller"
 
     choice
         prompt "TFT SPI Bus." if LV_TFT_DISPLAY_PROTOCOL_SPI
-        default LV_TFT_DISPLAY_SPI2_HOST
-        help
-              Select the SPI Bus the TFT Display is attached to.
+        default LV_TFT_DISPLAY_SPI2_HOST if LV_PREDEFINED_DISPLAY_TTGO && !IDF_TARGET_ESP32S2
```
arktrin: `LV_PREDEFINED_DISPLAY_TTGO && !IDF_TARGET_ESP32S2` also should always be true if `LV_PREDEFINED_DISPLAY_TTGO` is picked.

lvgl_tft/Kconfig
tore-espressif: SPI2 and SPI3 are identical, so I'd leave this section of code without change

arktrin: What is the purpose of this code anyway?

C47D: I can remove this change and leave only the fix on `LV_TFT_DISPLAY_X_OFFSET` and `LV_TFT_DISPLAY_Y_OFFSET` configs. What do you think?

C47D:
CI fails because of the missing fixes in the example, so #161 needs to be merged into develop and this PR updated in order to pass CI

lvgl_tft/Kconfig
C47D: How can the user set `IDF_TARGET_ESP32S2`? AFAIK only by using the `idf.py set-target` command. If so then we could remove the `&& !IDF_TARGET_ESP32S2` check from the if block, isn't?

arktrin: Yeah, I also think `&& !IDF_TARGET_ESP32S2` part should be removed.

lvgl_tft/Kconfig
arktrin: I think `&& !IDF_TARGET_ESP32S2` part should be removed at least.

C47D:
@arktrin How does 0852f39 look?

arktrin:
0852f39 looks ok. But what about esp32c3 where there is no SPI3_HOST ?

C47D:
82d956f should make the SPI3 not available for the ESP32C3

I think we should update the PR title

arktrin:
82d956f looks fine.

arktrin:
[52/53px offset issue is confirmed by spacewrench](https://github.com/lvgl/lv_port_esp32/issues/289#issuecomment-1005259086). So, this C code:
```
#elif (LV_HOR_RES_MAX == 240) && (LV_VER_RES_MAX == 135)
    #if (CONFIG_LV_DISPLAY_ORIENTATION_PORTRAIT) || \
        (CONFIG_LV_DISPLAY_ORIENTATION_PORTRAIT_INVERTED)
        offsetx1 += 40;
        offsetx2 += 40;
        offsety1 += 53;
        offsety2 += 53;
    #endif
#elif (LV_HOR_RES_MAX == 135) && (LV_VER_RES_MAX == 240)
    #if (CONFIG_LV_DISPLAY_ORIENTATION_LANDSCAPE) || \
        (CONFIG_LV_DISPLAY_ORIENTATION_LANDSCAPE_INVERTED)
        offsetx1 += 52;
        offsetx2 += 52;
        offsety1 += 40;
        offsety2 += 40;
    #endif
#endif
```
should be somehow adopted in Kconfig for `LV_PREDEFINED_DISPLAY_TTGO`. Maybe not in this PR. Maybe even after moving to lvgl v8, which is more important.

tore-espressif:
> should be somehow adopted in Kconfig for LV_PREDEFINED_DISPLAY_TTGO. Maybe not in this PR. Maybe even after moving to lvgl v8, which is more important.

The thing with display offsets is, that it can be different between eval-boards even if they have the same display controller and the same display resolution (they stay the same only with specific HW revision of the display, but the HW revision often changes in eval boards).

So after we change this hard-coded offset, few weeks later someone else will publish PR to change the offsets again for his particular case (HW revision).

It also makes it difficult to support other, non-standard resolutions, that require yet another offset settings.

So eventually it will have to be made runtime configurable, but I agree that it should be done in other PR :)

C47D:
Created #163 to keep track of it.

Can we merge this PR?

PR #180, opened 2022-02-10
Proposal: Display interface abstraction

C47D:
This is the current proposal of the display interface abstraction. The functions declared in `lvgl_tft/display_port.h` are meant to be implemented by the user. In this repo they are defined in `lv_port_display_espressif.c` in the `lv_port` directory, as this is the directory were all of the ESP-IDF will end up.

The `lv_disp_drv_t *drv` parameter is a pointer to the registered `lv_disp_drv_t`, so the user will be able to know what display should they handle, in this repo we currently only use one, so there's no use for it.

In a following commit I will use this interface in two display drivers, one using SPI and the other one using I2C, so we can see the abstraction works for both kind of interfaces.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,6 @@ build
 # Kconfig files
 sdkconfig
 sdkconfig.old
+
+# pre-commit configuration file
+.pre-commit-config.yaml
--- a/lv_port/lv_port_display_espressif.c
+++ b/lv_port/lv_port_display_espressif.c
@@ -17,7 +17,14 @@
 /* FIXME: Be able to get display driver I2C address from Kconfig */
 #if defined (CONFIG_LV_TFT_DISPLAY_CONTROLLER_SSD1306)
 #define OLED_I2C_ADDRESS    0x3C
+#else
+#define OLED_I2C_ADDRESS    0x00
 #endif
+
+// Control byte
+#define OLED_CONTROL_BYTE_CMD_SINGLE        0x80
+#define OLED_CONTROL_BYTE_CMD_STREAM        0x00
+#define OLED_CONTROL_BYTE_DATA_STREAM       0x40
 
 #define LV_DISPLAY_DC_CMD_MODE    0
 #define LV_DISPLAY_DC_DATA_MODE   1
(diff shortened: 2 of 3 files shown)


lv_port/lv_port_display_espressif.c
```diff
@@ -55,3 +85,77 @@ bool display_port_gpio_is_busy(lv_disp_drv_t *drv)
...
+        }
+    }
+
+    if (display_interface_is_i2c(drv)) {
+        uint8_t *data = (uint8_t *) args;
+        lvgl_i2c_write(OLED_I2C_PORT, OLED_I2C_ADDRESS, cmd, data, args_len);
+    }
+}
+
+void display_interface_send_data(lv_disp_drv_t *drv, void *data, size_t len)
```
kisvegabor, line 121: Finally the 
```c
void display_interface_send_data(lv_disp_drv_t *drv, lv_area_t * area,  void *data, size_t len)
```
prototype we discussed earlier won't be used here?

.gitignore
```diff
@@ -60,3 +60,6 @@ build
 # Kconfig files
 sdkconfig
 sdkconfig.old
+
+# pre-commit configuration file
+.pre-commit-config.yaml
```
kisvegabor, line 65: I think it's fine to add it and we can even promote it's usage.

lv_port/lv_port_display_espressif.c
C47D: To be honest I don't know, the way the drivers currently send the area information is very different to each other. I'm finishing porting a couple of drivers to use this interface to look how similar/different they end up being.

kisvegabor: Okay!

tore-espressif:
I hope this PR is not in a big rush. We finally got more manpower to this interface design and we'd like to test it on several displays and interfaces. I'll keep you updated, thanks!

C47D:
@tore-espressif it is not, hopefully we are going in the right direction.

.gitignore
C47D: I will upload the configuration file then 😄

PR #181, opened 2022-02-11
Allow user to configure custom SPI mode

### LINKED ISSUE #178: ST7789 SPI mode seems wrong
I have a display that is not working with the driver in SPI mode 2, but when changing the mode to 0, it works.

Not sure why the following code is needed in the first place.

#if defined (CONFIG_LV_TFT_DISPLAY_CONTROLLER_ST7789)
#define SPI_TFT_SPI_MODE    (2)
#else
#define SPI_TFT_SPI_MODE    (0)
#endif

C47D:
Fixes #178


C47D:
Merging as this doesn't change current behavior

PR #184, opened 2022-02-22
Initial changes to support ESP-IDF v5.0.0

rashedtalukder:
ESP-IDF v5.0.0 incorporates a major update to the real-time kernel and number of other changes. This is my quick stab at adding support for the new toolchain version before the official release. Tested on the M5Stack Core2 for AWS IoT EduKit and works great.

There are memory and performance advantages to using `pdMS_TO_TICKS` instead of `portTICK_PERIOD_MS` or the deprecated `portTICK_RATE_MS` macros. That's why you'll see mostly universal changes that switches to that.

### DIFF
--- a/lvgl_tft/esp_lcd_backlight.c
+++ b/lvgl_tft/esp_lcd_backlight.c
@@ -11,6 +11,23 @@
 #include "driver/gpio.h"
 #include "esp_log.h"
 #include "soc/ledc_periph.h" // to invert LEDC output on IDF version < v4.3
+#include "esp_idf_version.h"
+
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 0)
+#include "soc/gpio_sig_map.h" // 5.0
+#if defined (CONFIG_IDF_TARGET_ESP32)
+#include "esp32/rom/gpio.h"
+#endif
+#if defined (CONFIG_IDF_TARGET_ESP32S2)
+#include "esp32s2/rom/gpio.h"
+#endif
+#if defined (CONFIG_IDF_TARGET_ESP32S3)
+#include "esp32s3/rom/gpio.h"
+#endif
+#if defined (CONFIG_IDF_TARGET_ESP32C3)
+#include "esp32c3/rom/gpio.h"
+#endif
+#endif
 
 typedef struct {
     bool pwm_control; // true: LEDC is used, false: GPIO is used
--- a/lvgl_tft/esp_lcd_backlight.c
+++ b/lvgl_tft/esp_lcd_backlight.c
@@ -9,24 +9,13 @@
 #include "esp_lcd_backlight.h"
 #include "driver/ledc.h"
 #include "driver/gpio.h"
+#include "rom/gpio.h"
 #include "esp_log.h"
 #include "soc/ledc_periph.h" // to invert LEDC output on IDF version < v4.3
 #include "esp_idf_version.h"
 
 #if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 0)
-#include "soc/gpio_sig_map.h" // 5.0
-#if defined (CONFIG_IDF_TARGET_ESP32)
-#include "esp32/rom/gpio.h"
-#endif
-#if defined (CONFIG_IDF_TARGET_ESP32S2)
-#include "esp32s2/rom/gpio.h"
-#endif
-#if defined (CONFIG_IDF_TARGET_ESP32S3)
-#include "esp32s3/rom/gpio.h"
-#endif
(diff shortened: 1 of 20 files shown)


C47D:
Hi, changes look OK, the only 'issue' I see are the rom gpio includes, they were removed in the develop branch. I will try to patch develop with the other ones.

rashedtalukder:
Looks like you have to include the target specific ROM libraries as per: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/migration-guides/system.html

It threw an undefined error without it.

Rashed Talukder

lvgl_tft/esp_lcd_backlight.c
```diff
@@ -11,6 +11,23 @@
...
+#endif
+#if defined (CONFIG_IDF_TARGET_ESP32S2)
+#include "esp32s2/rom/gpio.h"
+#endif
+#if defined (CONFIG_IDF_TARGET_ESP32S3)
+#include "esp32s3/rom/gpio.h"
+#endif
+#if defined (CONFIG_IDF_TARGET_ESP32C3)
+#include "esp32c3/rom/gpio.h"
+#endif
```
tore-espressif: Could you please try simply:
```suggestion
#include "rom/gpio.h"
```
?
CMake should include the correct file for you

lvgl_tft/ili9341.c
```diff
@@ -9,6 +9,7 @@
 #include "ili9341.h"
 #include "disp_spi.h"
 #include "driver/gpio.h"
+#include "esp32/rom/gpio.h"
```
tore-espressif: Same here

tore-espressif (changes requested):
Thank you very much for the PR. Could you please try the suggested change? Thanks!

lvgl_tft/esp_lcd_backlight.c
rashedtalukder: I should've just tried that originally or checked. The ESP-IDF doc didn't allude to this.

lvgl_tft/ili9341.c
rashedtalukder: Updated all the files with references to the undefined variable. Compiled without error and works.

tore-espressif (approved):

ropg:
I haven't had time to look at ESP-IDF 5. Wouldn't these changes break under IDF 4? Maybe it would be best to create our own macro and use `#IFDEF` to interpret differently based on ESP-IDP version?

rashedtalukder:
> I haven't had time to look at ESP-IDF 5. Wouldn't these changes break under IDF 4? Maybe it would be best to create our own macro and use `#IFDEF` to interpret differently based on ESP-IDP version?

These are backwards compatible changes. I tested down to IDF 4.2.1 and had no issue. The one header include that is only available in v5+ has a `#IFDEF` to prevent breakage: https://github.com/lvgl/lvgl_esp32_drivers/blob/a75abc016e99bf757212cf31de4ecc9645ad9abe/lvgl_tft/esp_lcd_backlight.c#L17

tsctrl:
> These are backwards compatible changes. I tested down to IDF 4.2.1 and had no issue. The one header include that is only available in v5+ has a `#IFDEF` to prevent breakage:

it is not compatible, i use release v4.4.0, build failed and lots of files and declaration are missing which only available in v5.

rashedtalukder:
> > These are backwards compatible changes. I tested down to IDF 4.2.1 and had no issue. The one header include that is only available in v5+ has a `#IFDEF` to prevent breakage:
> 
> it is not compatible, i use release v4.4.0, build failed and lots of files and declaration are missing which only available in v5.

What were your errors?

tsctrl:
mostly on spi access function. probably c linkage or compiler causing the issue. it is working after switching branch. i am not able to reproduce the issue now. will try again later.

edit: probably user have to add v5 support to their local idf installation to make it work with v4. that what i did that solve the build issue. 

thanks!

rashedtalukder:
You might be seeing the effects of there being different branches that have different commits. I couldn't get the Master branch to work without other changes that were done in the develop branch.

Glad this PR does not have changes that result in errors for you though.

rashedtalukder:
@tsctrl, I also pushed the changes to my fork's develop branch https://github.com/rashedtalukder/lvgl_esp32_drivers/commit/44342b653f4111cb5340a93338b553f120898809 which should be better for a future merge since it seems that the develop branch is most up to date.

@tore-espressif @C47D if you'd like me to submit a new PR for the develop branch and close this one, let me know.

C47D:
Hi @rashedtalukder, I think that a PR against develop is preferred. Sorry for not following along the discussion, I've been somewhat busy with other stuff.

rashedtalukder:
Closing to open a new PR on the develop branch instead.

PR #198, opened 2022-05-13
Merge develop into master branch

C47D:
Hi @tore-espressif, I've been busy with day job, but I think we should be able to merge develop into master.


tore-espressif:
@C47D Looks good to me.

Just a side note: We released `esp_lcd_touch` component few days ago, that should work in tandem with `esp_lcd`, primitive example here: https://github.com/espressif/esp-bsp/tree/master/examples/touchscreen_colorwheel

I'll comment into LVGL's driver refactor discussion soon ;)

shymega:
I've noted that the EVE errors that I'm experiencing with #213 are resolved with the `develop` branch, but then there are other unrelated errors. I would really like for `develop` and mainline to be more in sync, as there's a huge difference between them now, and with LVGL v8 out now, it feels like it needs to be sorted.

Is there any update on this issue? The deviation between these branches is substantial, and there are a lot of PRs too. I can help out if you need me to - I'm working on an ESP32 here locally with a supported TFT display controller and touch panel.

PR #203, opened 2022-08-14
cst816 touch driver

abolbis:
I implemented cst816 touch driver https://github.com/fbiego/dt78/blob/master/datasheets/CST816S.pdf
This has been tested on an ESP32-S3-WROOM-1 module. While testing I noticed a bugs related to SPI DMA and fixed it. 
I know it would have been better to branch from develop but I noticed that there are major changes between develop and master and I started implementing all my work from master.


krupis:
Hello. I use T-Display S3 module:
https://www.lilygo.cc/products/t-display-s3

I got the basic example working based on i80_controller esp-idf example. My example project can be found here:
https://github.com/krupis/T-Display-S3-esp-idf


According to T-Display-S3 documentation, the display uses CST816 touch driver:
image

But in the menuconfig, I can only choose from the following:
image


Could you please help me understand how can I use the CST816 driver for my project?

PR #204, opened 2022-08-19
Feature/product build

skip-dt:
This change is needed to get lvgl to compile the esp32s3 processor used on the product board.
Background:  several drivers (which we don't use) won't compile unless CONFIG_LV_DISP_PIN_BUSY is
defined.


skip-dt:
Sorry, wrong rep

Issue #115, opened 2021-09-21
kconfig for lvgl_esp32_drivers/lvgl_i2 causes ESP-IDF SDK Configuration Editor to Crash


wreyford:
I've updated to the latest master of lvgl_esp32_drivers. 
I'm using LVGL8.3 master.
![error_I2C_Port](https://user-images.githubusercontent.com/50998190/134126987-d4a94046-af07-4d6d-b57c-19c49837f02b.jpg)
I can select i2C port Zero, but in the KConfig specific to I2C Port Settings, 
if I select Enable I2C port 0 or Enable I2C port 1, instead of bringing up the options to set the pins etc, it just crashes menuconfig.

Please, I don't know how to set my pins.

Also, when including this line of code as per the readme before calling lv_init():
 ``` 
lvgl_i2c_locking(i2c_manager_locking()); 
```

I get the following compiler error:
> implicit declaration of function 'i2c_manager_locking'; did you mean 'i2c_master_write'? [-Werror=implicit-function-declaration]

wreyford:
In the sdkconfig, the following [ CONFIG_LV_I2C=y and CONFIG_LV_I2C_DISPLAY_PORT=0] appears to be set, but it is not selected anywhere in menuconfig:

'''
# CONFIG_LV_DISP_BACKLIGHT_OFF is not set
CONFIG_LV_DISP_BACKLIGHT_SWITCH=y
# CONFIG_LV_DISP_BACKLIGHT_PWM is not set
# CONFIG_LV_BACKLIGHT_ACTIVE_LVL is not set
CONFIG_LV_DISP_PIN_BCKL=-1
CONFIG_LV_I2C=y
CONFIG_LV_I2C_DISPLAY_PORT=0
# end of LVGL TFT Display controller
'''

and in the kconfig file for lvgl_esp32_drivers/lvgl_tft/kconfig:

line 996
'''
config LV_I2C
        bool
        default y if LV_I2C_DISPLAY

    config LV_I2C_DISPLAY_PORT
        int
        default 1 if LV_I2C_DISPLAY_PORT_1
        default 0
'''

And  at line
177

'''
 # Display controller communication protocol
    #
    # This symbols define the communication protocol used by the
    # ESP32 to communicate with the display controller.
    # This symbols can be used, but not limited, to:
    # - Know what peripherals to initialize.
    # - Know if the touch and display controllers share the same peripheral.
    # - Etc.
[...]

PR #214, opened 2022-12-22
Fix ESP-IDF configuration server error when enabling I2C ports 0 and 1

mateusglucas:
Fix ESP-IDF configuration server error when enabling I2C ports 0 and 1 (#115). The previous fix (#132) only fixed the problem for I2C port 0.


vleppavuori (commented):
This is definitely needed.  Did same change in own development. Removes crash when trying to edit i2c settings

vleppavuori (approved):

